PR #10, opened 2023-10-20
Allow secrets to be managed by an external provider so the chart values can be saved in git

deanevansworldline:
Currently the databaseConnection string is a hard requirement.
For systems like Flux that requires the helm values to be stored in git, this is a problem as the database credentials would need to be in git.

This new option will allow the chart to install but not to manage the secrets. This effectively allows the secrets to be managed by third party operators like sealed-secrets or external-secrets.


APErebus (approved):
Thanks for the PR!

That seems like a good way to handle it :)

PR #11, opened 2023-11-21
First pass of a helm chart for kubernetes tentacle

scme0:
# Background

Part of #project-k8s-agent

We are adding a new helm chart to be used by customers wanting to install the new Kubernetes Tentacle. This helm chart will be used by an install script which will be generated by a wizard in octopus when users request to install a k8s target. This will make the script more succinct and make it easier for customers to manage.

# Results

This is the first pass for the helm chart. It currently works for Polling tentacles only with an input close to this:
Screenshot 2023-11-21 at 12 23 17
(It also works with serverApiKey in place of bearer token)

After this is in, we can iterate on it and add functionality. The next thing to add will be to create the appropriate rules for the service account so that it can issue jobs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,6 @@
 
 This repository contains the officially supported [Helm](https://helm.sh) chart for deploying [Octopus Deploy](https://octopus.com).
 
-- [Octopus Deploy](https://github.com/OctopusDeploy/helm-charts/tree/main/charts/octopus-deploy). Self-host the Octopus Deploy server. 
-- Workers. _Coming Soon_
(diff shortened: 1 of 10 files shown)


scme0:
@APErebus As discussed, I've added and updated the READMEs. Does this draw enough attention?

Screenshot 2023-11-23 at 10 01 28

This is the main README (Should I put the Caution there too?)

Screenshot 2023-11-23 at 10 01 28

APErebus (approved):
@scme0 I tweaked the readme to be a bit more clear that it's in development and should not be used.

Apart from that, great start!

PR #13, opened 2023-12-01
Significant changes to the Kubernetes Tentacle chart

APErebus:
This PR contains a number of things that bring this up to a "working" Alpha version

- Adds a separate NFS deployment + service for the NFS container that's needed. This has to run in a separate deployment because it's used as a volume in the tentacle pod and if its included in that pod (as a sidecar), the pod doesn't start
- Add service account roles and bindings for the tentacle service account and the job service account
- Add a ton of environment variables to tentacle container
- Remove a bunch of configuration we don't want to allow just yet

I feel like there is a bunch of naming cleanup we should do here, but this is a good start

Shortcut story: [sc-60399]


MissedTheMark (approved):
Changes look good to me 👍

PR #16, opened 2023-12-15
Tentacle Service Account now can access pods API

APErebus:
This is needed to support script cancellation in  https://github.com/OctopusDeploy/OctopusTentacle/pull/742


APErebus:
No longer needed!

PR #20, opened 2023-12-18
Add changesets/cli support for versioning Kubernetes Tentacle chart

APErebus:
The Kubernetes Tentacle is now version via [changesets](https://github.com/changesets/changesets).

When adding a new change, if it requires a version bump, then `npm run changeset` should be done from the `kubernetes-tentacle` folder. This will propt to add a major/minor/patch version + message.

Include the generated file in the PR and then when it's merged, the new workflow _should_ create a new PR with the versioning which will update the `package.json` and the `Chart.json` with the correct version


.github/workflows/create-versioning-pr.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+    - run: |
+        version=$(jq -r .version ./charts/kubernetes-tentacle/package.json)
+        version="$version" yq -i '.version = strenv(version)' ./charts/kubernetes-tentacle/Chart.yaml
+        git config user.name github-actions
+        git config user.email github-actions@github.com
+        git add ./charts/kubernetes-tentacle/Chart.yaml
+        git commit -m "Update Chart.yaml"
+        git push
+      if: steps.changesets.outputs.hasChangesets == 'true'
```
APErebus, line 47: this little script is used to update the `version` property of the `Chart.yaml` based on the `package.json` `version` property then commit it back to the PR that was just opened 🤞

MissedTheMark (approved):
Very nice!

PR #21, opened 2023-12-18
Fix versioning workflow

APErebus:
Turns out having an invalid workflow doesn't give you a red build if that workflow doesn't run :(


MissedTheMark (approved):
Derp to Un-derp 🎉

PR #26, opened 2024-01-11
Change to support config map data for tentacle registration

MissedTheMark:
# Background

Part of #project-k8s-agent

# Details

We need to allow the Tentacle to read and edit config maps so that it can save it's configuration to a config map.

We need to save the Tentacle's configuration in a config map to ensure that it's able to recover if either the NFS volume or the tentacle container goes down and comes back up. This can happen at any time in Kubernetes due to load balancing.

### DIFF
--- a/charts/kubernetes-tentacle/.changeset/violet-toes-wait.md
+++ b/charts/kubernetes-tentacle/.changeset/violet-toes-wait.md
@@ -1,0 +1,5 @@
+---
+"kubernetes-tentacle": patch
+---
+
+Allow Tentacle service account to interact with config maps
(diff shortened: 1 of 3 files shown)


scme0:
[sc-66386]

APErebus (approved):
Yay! Hopefully the versioning workflow works

scme0:
@APErebus Yeah the changesets comment didn't update to say what version changes there are so I'm a bit unsure if it has worked correctly... I guess I can try it and if it doesn't work we can fix it. I'm almost done testing - ran into a few unrelated issues which I'm fixing.

APErebus:
The changeset version should only update when you merge to `main`

PR #34, opened 2024-01-15
Additional k8s tentacle updates to support configuration stored in config map

scme0:
# Background

This is part of #project-k8s-agent. We are updating the agent to save all it's configuration in a config map to ensure that it's not lost if either the nfs or the tentacle pod goes down.

# Changes

1. I've added in an empty config map to the helm chart so that if you uninstall the tentacle via helm, it will clean up the configmap. Previously it would leave it behind because it was created by the tentacle itself and not by helm.

2. I've also added in an `InstanceName` env var because this is used to get the name of the config map. This means that each k8s tentacle in an instance needs to have a unique instance name, we just pass in the 'fullname' from the helm chart which is unique for targets as it's based off the target name. This change will coincide with changes made to the install script in Tentacle.

3. Lastly I've added the ability to specify if you want to use the arm nfs container. This is in a new "developer" object as it's only really required when running locally on an M1 mac or other arm based development machine.

### DIFF
--- a/charts/kubernetes-tentacle/templates/configmap.yaml
+++ b/charts/kubernetes-tentacle/templates/configmap.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: "{{ include "kubernetes-tentacle.fullname" . }}-configmap"
--- a/charts/kubernetes-tentacle/templates/configmap.yaml
+++ b/charts/kubernetes-tentacle/templates/configmap.yaml
@@ -1,4 +1,5 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: "{{ include "kubernetes-tentacle.fullname" . }}-configmap"
+  name: tentacle-config
+  namespace: {{ .Release.Namespace | quote }}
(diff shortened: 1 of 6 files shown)


charts/kubernetes-tentacle/templates/configmap.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: "{{ include "kubernetes-tentacle.fullname" . }}-configmap"
```
APErebus, line 4: missing the `namespace`

APErebus (changes requested):
The config map should be namespace'd correctly. This means we don't need to worry about custom instance names because everything will be fully isolated within the namespace

charts/kubernetes-tentacle/templates/configmap.yaml
APErebus, line 4:
```suggestion
  name: "{{ include "kubernetes-tentacle.fullname" . }}-config"
```

scme0: Ah good call

scme0:
@APErebus That's true, good point. I can revert it back to use a fixed 'tentacle-config' then with the unique namespace 👍

APErebus:
> @APErebus That's true, good point. I can revert it back to use a fixed name: 'tentacle-config' then with the unique namespace 👍

I'm assuming it uses the instance name to build the config map name? If so, then yeah, I think that's fine

APErebus (approved):
very good!

PR #43, opened 2024-01-22
Rename kubernetes tentacle to agent

scme0:
We have decided to rename the tentacle to an agent as the tentacle component is only part of the complete agent.


charts/kubernetes-agent/.changeset/spicy-paws-float.md
```diff
@@ -0,0 +1,5 @@
+---
+"kubernetes-agent": patch
```
APErebus, line 2: let's go minor here, so we get a meaningful-ish version bump

charts/kubernetes-agent/values.yaml
```diff
@@ -1,11 +1,11 @@
-# Default values for kubernetes-tentacle.
+# Default values for kubernetes-agent.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
 replicaCount: 1
 
 image:
-  repository: docker.packages.octopushq.com/octopusdeploy/kubernetes-tentacle
+  repository: docker.packages.octopushq.com/octopusdeploy/kubernetes-agent
```
APErebus, line 8: this won't change as the tentacle image will still be `tentacle`

APErebus (changes requested):

charts/kubernetes-agent/values.yaml
scme0: Damn, good catch!

APErebus (approved):

scme0:
[sc-69136]

PR #45, opened 2024-01-22
Rename and re-organize resources

APErebus:
Makes some significant changes to the names of resources.

The key change is the "base" name is now `octopus-agent`, rather than `.Chart.Name` (`kubernetes-tentacle`).

We also now don't have a "fullname", which included the release name. This seemed pointless because as almost every resources is in the namespace, then the extra release name wasn't helping.
The only scenario we have this is is for cluster-wide resources (cluster role & cluster role bindings)

I also just cleaned up filenames, split files to one resource per-file and general cleanup


charts/kubernetes-agent/package-lock.json
APErebus, line 1: I need to update the versioning workflow to also update the `package-lock.json` so it doesn't keep being added to PR's

scme0 (approved):
LGTM 🚀

charts/kubernetes-agent/package-lock.json
APErebus: I did this in #46

Issue #23, opened 2023-12-22
Make AcceptEula easier to work with


trondhindenes:
the chart takes the octopus.acceptEula and uses it like this: !!str "{{.Values.octopus.acceptEula}}". It is tricky to programmatically generate the values file, since yaml by default translates `Y` to `true`. Since the chart string-parses this value, one has to go thru a lot of hoops in order to construct the required value without having it turned into `true`if run thru a yaml parser/generator. It would be much better if this chart implemented a template function that accepted a string OR a boolean, and converted that into the `Y` that seems to be required by Octopus Deploy. 

....Or just get rid of this as a required value altogether and instead just have a waiver that "by installing this chart you accept the EULA". Much simpler imho.

PR #50, opened 2024-01-30
allow supply bools or strings for accepting the EULA

MJRichardson:
closes https://github.com/OctopusDeploy/helm-charts/issues/23


APErebus (approved):
noice! We might steal that for the kubernetes agent

PR #51, opened 2024-01-31
Use new multi-arch tentacle image

scme0:
# Background

Part of #project-k8s-agent.

[I've recently updated](https://github.com/OctopusDeploy/OctopusTentacle/pull/781) the way we build the kubernetes tentacle container image, so now it builds as a multi-arch image. This means that we don't need to postfix the container tag with `-linux-{arch}`.

# Results

I've removed the postfix from the tentacle image tag and I've updated the tentacle version to the first version which will support the new multi-arch image.


APErebus (approved):
woo!

PR #55, opened 2024-02-07
Specify Linux as the target nodes

APErebus:
We only support running on Linux, so force the node selection for the two deployments

Shortcut story: [sc-70536]


scme0 (approved):
🚀

PR #57, opened 2024-02-07
Add resource requests for NFS & Agent containers

APErebus:
To be good Kubernetes citizens, we need to set resource requests for our containers.

Based on testing on an AKS cluster, these limits seem quite reasonable

Shortcut story: [sc-70554]

### CPU
image

### Memory
image

## CPU

image

## Memory

image


APErebus:
I will update this PR with the Tentacle container that contains the job resource requests

flin-8 (approved):
LGTM

PR #60, opened 2024-02-09
Fix Kubernetes Agent publish workflow to not overwrite existing versions

APErebus:
There was an issue where the Kubernetes Agent publish workflow would publish a version of `main` with the _new` chart, but using the `previous` version (overwriting the old version).

This PR updates the logic so that after versioning & packaging, we publish with these rules:

Artifactory: All non-main/pre-releases and also `main` where the commit _is_ a versioning commit
DockerHub: `Main` only w where the commit _is_ a versioning commit

This means that when a PR is merged that isn't a versioning commit, that there is no publishing of that version


scme0 (approved):
LGTM 👍

PR #61, opened 2024-02-09
Change to use DockerHub hosted images

APErebus:
Now we are publishing both the nfs-server and kubernetes-tentacle containers to dockerhub, so lets use them!


scme0 (approved):
🚀

PR #64, opened 2024-02-14
Delay shutting down the NFS Pod so the Tentacle Pod doesn't hang while umount'ing the fileshare

flin-8:
[sc-70770]

We discovered that if the NFS Pod shuts down before the Tentacle Pod does, then the Tentacle Pod will be stuck in a `terminating` state because kubelet can't `umount` the NFS share. This will affect users when trying to uninstall the Kubernetes Agent.

Our solution is to delay shutdown of the NFS Pod such that the Tentacle Pod shuts down first.
 
The solution is implemented as a prestop hook, which will keep waiting as long as there's an active NFS connection:
CleanShot 2024-02-14 at 14 38 49@2x

The termination grace period is set at 2min (120s), which will forcibly shut down the NFS Pod even if there are active connections.

Prestop hook initiated - waiting for NFS connections to close:
CleanShot 2024-02-14 at 14 52 23@2x

Tentacle Pod has shut down, NFS Pod shuts down:
CleanShot 2024-02-14 at 14 53 31@2x

Tentacle Pod hasn't shut down in time, NFS Pod is forced to shut down after grace period:
CleanShot 2024-02-14 at 15 10 07@2x


charts/kubernetes-agent/templates/nfs-deployment.yaml
```diff
@@ -26,6 +26,20 @@ spec:
...
           image: "octopusdeploy/nfs-server:latest"
+          lifecycle:
+            preStop:
+              exec:
+                command:
+                  - /bin/bash
+                  - '-c'
+                  - >-
+                    echo "Prestop: Delaying shutdown until NFS connections are
+                    closed" > /proc/1/fd/1;  counter=0;  nfsConnection=$(netstat
```
flin-8, line 37: redirecting the output shows these message in the Pod logs - is this bad?

scme0: On the contrary, that's good isn't it?

scme0 (approved):
Nice solution 👍

.changeset/poor-windows-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
```
APErebus, line 2: what happened here 😕

charts/kubernetes-agent/templates/nfs-deployment.yaml
APErebus: I'd say that's good 👍

APErebus (changes requested):
I think the changeset won't work as it's the wrong package

.changeset/poor-windows-cover.md
scme0: Oh good catch! 🤔

flin-8: I just used the generated default value and left it for you to catch 😄

.changeset/poor-windows-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+"kubernetes-agent": patch
```
flin-8, line 2: @APErebus better?

APErebus: much 🙇

APErebus (approved):

PR #65, opened 2024-02-14
Support defining custom job service account permissions

APErebus:
By default, the job service account gets a `ClusterRole` that can do anything (effectively cluster admin). This may be undesirable for certain customers.

This PR allows for the job service account permissions to be configured 2 different ways:

1. Replace the `rules` for the existing `ClusterRole`
2. Replace the `ClusterRole` and `ClusterRoleBinding` with namespaced `Role` and `RoleBinding`, allow for an agent to only perform jobs/actions on specific namespaces

The new `jobServiceAccount.clusterRole.rules` value node allows for swapping the default cluster role rules
The new `jobServiceAccount.role.rules` value node disables the cluster role and creates a namespaced role with the defined rules

Shortcut story: [sc-68264]


scme0 (approved):
Nice! Good to get some tests in there to start with!

PR #69, opened 2024-02-15
Update nfsPort to nfsIpAddress

scme0:
This value is an IP address, not a port so lets update the name to be consistent 👍


flin-8 (approved):
LGTM

PR #71, opened 2024-02-15
Update to use pnpm and changesets from root

scme0:


pnpm-workspace.yaml
```diff
@@ -0,0 +1,3 @@
+packages:
+  - 'charts/**'
+  - '!**/octopus-deploy'
```
APErebus, line 3: would this be removed when/if we change the octopus deploy helm chart to use versioning?

APErebus (approved):

pnpm-workspace.yaml
scme0: Yep, for now it's just to stop an unexpected behaviour but would be removed when there is a package.json in that folder.

PR #75, opened 2024-02-19
Update versions of github actions which were showing warnings

scme0:


scme0:
Closing because this is too much trouble for the moment

PR #78, opened 2024-02-19
Bump tentacle version

scme0:
This new version doesn't contain any significant changes for the k8s agent but I'll use it to test if the automatic updates work correctly.


APErebus (approved):
:shipit:

PR #80, opened 2024-02-19
Change to use only CSI driver for NFS

APErebus:
To remove the need to have fixed ClusterIP for the NFS container, this PR changes to use the [csi-driver-nfs](https://github.com/kubernetes-csi/csi-driver-nfs).

This allows us to reference the DNS address of the NFS container. We create a Persistent Volume & Persistent Volume Claim that uses the CSI driver as it's provisioner.

Fortunately, this is a non-breaking, in -place upgrade as existing agents will just use the previously static IP, but new agents will receive a dynamic IP.

To install the CSI driver, this is the Helm command to run

```
helm upgrade --install --atomic `
--repo https://raw.githubusercontent.com/kubernetes-csi/csi-driver-nfs/master/charts `
--namespace kube-system `
--version v4.6.0 `
--set externalSnapshotter.enabled=true `
--set controller.runOnControlPlane=true `
csi-driver-nfs `
csi-driver-nfs
```

Shortcut story: [sc-71123]


charts/kubernetes-agent/templates/_helpers.tpl
```diff
@@ -6,6 +6,11 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{ .Values.nameOverride | default "octopus-agent" | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
+{{- define "kubernetes-agent.fullName" -}}
```
APErebus, line 9: used for cluster-wide resources

charts/kubernetes-agent/templates/_helpers.tpl
```diff
@@ -6,6 +6,10 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+{{ (printf "%s-%s" (.Values.nameOverride | default "octopus-agent") .Release.Name) | trunc 63 | trimSuffix "-" }}
```
flin-8, line 10: Considering reusing `kubernetes-agent.name`

flin-8 (approved):
Are there any high value tests we could add?

APErebus:
> Are there any high value tests we could add?

I actually added the tests in the PV volume PR following it #81. Let me know if you want to see some more in there

scme0 (approved):
🚀

PR #81, opened 2024-02-20
Allow specifying PV instead of NFS server

APErebus:


charts/kubernetes-agent/templates/_helpers.tpl
```diff
@@ -80,13 +79,23 @@ The name of the secret to store the authentication information (bearer token/api
...
-{{- if .Values.storage.nfs.enabled }}
 volumes:
 - name: tentacle-home
   persistentVolumeClaim:
-    claimName: {{ include "nfs.pvcName" .}}
-{{- end }}
-{{- end }}
-
+    claimName: {{ include "kubernetes-agent.pvcName" . }}
+{{- end }}
```
APErebus, line 101: In a future PR, this could just change to passing the `claimName` directly, rather than the JSON

.github/workflows/kubernetes-agent-publish-chart.yaml
```diff
@@ -9,8 +9,6 @@ on:
     - .github/workflows/kubernetes-agent-publish-chart.yaml
 
   pull_request:
-    branches:
-    - '*'
```
APErebus, line 13: Allows the publish to run for PR's

scme0: Yeah this is a security issue isn't it?

APErebus: hrm, good question... I'll ask in #secops-requests...

APErebus: I can revert this while we wait for an answer

APErebus: reverted in 
[f592fe9](https://github.com/OctopusDeploy/helm-charts/pull/81/commits/f592fe92f8be5d9e45d0f2ec5d90f3f9e1b0d8e0)

scme0: Wait I think I'm confused, so removing the `*` allows it to run on branches? I thought `*` would apply to all branches? 🤔

Anyways, yeah I'm unsure if it's ok - if someone external has access to running github workflows, is that ok?

APErebus: correct, removing this means it runs on all branches. There is some weird thing going on with the `*` not behaving

scme0 (approved):
Nice work! Will this merge cleanly with the other PR for NFS?

APErebus:
> Nice work! Will this merge cleanly with the other PR for NFS?

I'll rebase this PR once that other one is merged. They'll go through as one release

PR #82, opened 2024-02-20
Add hello-octo test helm chart

scme0:
This test helm chart will be used in tests needing a helm chart in an oci repository.

I've updated the versioning workflow to be more generic so it produces a commit with title "Version Helm Charts". I've also added a workflow for the new chart so it publishes to artifactory. It doesn't publish to docker as it's only for tests.

### DIFF
--- a/.github/workflows/create-versioning-pr.yaml
+++ b/.github/workflows/create-versioning-pr.yaml
@@ -40,8 +40,8 @@ jobs:
       id: changesets
       uses: changesets/action@v1
       with:
-        commit: 'Version Kubernetes Agent Chart'
-        title: 'Version Kubernetes Agent Chart'
+        commit: 'Version Helm Charts'
+        title: 'Version Helm Charts'
         createGitHubReleases: false
       env:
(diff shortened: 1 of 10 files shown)


.github/workflows/create-versioning-pr.yaml
APErebus, line 1: We need to adjust the paths for when this executes

scme0: Good call, I've added the hello-octo folder 👍

APErebus, line 1: Just thinking about this more. Do we want a single PR to be able version multiple charts, or is that just how it works the the changeset action?🤔

scme0: Typically a mono-repo has sub projects which are related and potentially reference each other. This is a bit of a weird use case as we have 2 (and soon to be three) which are independent but are using changeset. I don't think it matters too much though if they are all versioned in the same commit? You wouldn't really ever want to undo a versioning commit in normal usage would you?

scme0:
I'm not going to merge this for now as I don't think it's required anymore. I'll keep it open for the moment until I'm sure.

PR #83, opened 2024-02-22
Change to use Pods, not Jobs

APErebus:
Updates various files & names to support the change in https://github.com/OctopusDeploy/OctopusTentacle/pull/823 where script execution is performed via pods, not jobs

Shortcut story: [sc-71684]


scme0 (approved):
LGTM

PR #86, opened 2024-03-08
Add default namespace as option for tentacle deployment

scme0:
#project-k8s-agent

See https://github.com/OctopusDeploy/OctopusDeploy/pull/23271 for details

[sc-71237]

### DIFF
--- a/charts/kubernetes-agent/templates/tentacle-deployment.yaml
+++ b/charts/kubernetes-agent/templates/tentacle-deployment.yaml
@@ -58,6 +58,8 @@ spec:
               value: {{ join "," .Values.tentacle.targetEnvironments | quote }}
             - name: "TargetRole"
               value: {{ join "," .Values.tentacle.targetRoles | quote }}
+            - name: "DefaultNamespace"
+              value: {{ .Values.tentacle.defaultNamespace | quote }}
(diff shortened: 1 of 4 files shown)


charts/kubernetes-agent/templates/tentacle-deployment.yaml
```diff
@@ -58,6 +58,8 @@ spec:
               value: {{ join "," .Values.tentacle.targetEnvironments | quote }}
             - name: "TargetRole"
               value: {{ join "," .Values.tentacle.targetRoles | quote }}
+            - name: "DefaultNamespace"
```
flin-8, line 61: We talk about renaming these properties to say "InitialXYZ" at one stage, did we not action it?

scme0: Oh I don't remember this, but it makes sense to me and it makes it clearer that this might not be up-to-date with what's in server.

Is it better to do that in a separate PR? (I can change the default namespace one in this PR, and do the ther properties in a separate PR).

flin-8: Yeah separate PR is fine - or if you want to leave this one like this and rename them all together

flin-8 (approved):
LGTM

PR #88, opened 2024-03-14
Fix bug where pod monitoring stops working after a while

flin-8:
[sc-73237]
This updates the Tentacle image to pick up https://github.com/OctopusDeploy/OctopusTentacle/pull/836


scme0 (approved):
🚀

PR #92, opened 2024-03-18
Move nfs container info to values and use new versioned tag (instead of latest)

scme0:
Now we can start versioning the nfs server image when we make different changes to it. 👍


APErebus (approved):
Awesome!

PR #93, opened 2024-03-18
Reduce termination grace period so things clean up quicker in the worst case

scme0:
Does this make sense to you? I feel like 30 seconds should be enough? 🤔


APErebus (approved):
yeah, that's more reasonable 👍

PR #99, opened 2024-04-02
Make NFS deployment use default storage class instead of empty dir

scme0:
When we use the default storage class, if the nfs pod restarts on the same node, the data on the nfs is not lost and the pods with nfs mounts can recover. We use the default storage class by default but allow users to define their own storage class if they want.

[sc-74270]

### DIFF
--- a/charts/kubernetes-agent/templates/nfs-pod-pvc.yml
+++ b/charts/kubernetes-agent/templates/nfs-pod-pvc.yml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
-  name: {{ include "nfs.nfsPodPvcName" . }}
+  name: {{ include "nfs.podPvcName" . }}
 spec:
   accessModes:
     - ReadWriteOnce
--- a/charts/kubernetes-agent/templates/_nfs-helpers.tpl
+++ b/charts/kubernetes-agent/templates/_nfs-helpers.tpl
@@ -14,7 +14,7 @@ These are used for the NFS container & resources
 {{- printf "%s-pv" (include "nfs.fullName" .) }}
 {{- end }}
 
-{{- define "nfs.nfsPodPvcName"}}
+{{- define "nfs.podPvcName"}}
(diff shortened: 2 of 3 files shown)


charts/kubernetes-agent/templates/nfs-pod-pvc.yml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ include "nfs.podPvcName" . }}
+spec:
+  accessModes:
+    - ReadWriteOnce
+{{- if not (quote .Values.persistence.nfs.podStorageClassName | empty) }}
+  storageClassName: {{quote .Values.persistence.nfs.podStorageClassName }}
+{{- end }}
```
APErebus, line 10: this allows them to use storageClassName="" and have it work?

scme0: Yep, I've tested this and this:
```
helm template \
--set persistence.nfs.podStorageClassName="" \
--set tentacle.logLevel="Trace" \
--set tentacle.ACCEPT_EULA="Y" \
--set tentacle.targetName="scme" \
--set tentacle.serverUrl="http://host.docker.internal:8066/" \
--set tentacle.serverCommsAddress="http://host.docker.internal:10943/" \
--set tentacle.space="Default" \
--set tentacle.targetEnvironments="{dev}" \
--set tentacle.targetRoles="{scme-role}" \
--set tentacle.serverApiKey="API-LRJCU6XSMGQQCLC2W5UQQOZGCH6Z06T" \
--create-namespace --namespace octopus-agent-scme \
scme \
/Users/scottmerchant/Documents/dev/helm-charts/charts/kubernetes-agent
```
comes back including this:
```
# Source: kubernetes-agent/templates/nfs-pod-pvc.yml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: octopus-agent-nfs-scme-pod-pvc
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: ""
  resources:
    requests:
      storage: 1Gi
```

APErebus (approved):

PR #100, opened 2024-04-02
Use NFS mount timeout instead of getting the nfs pod to wait for connections to close before shutting down

scme0:
# Background

When we add support to use the default storage class as the basis for the nfs pod (https://github.com/OctopusDeploy/helm-charts/pull/99), trying to access a broken nfs mount on tentacle hangs instead of just failing with a stale file handle. To fix this, we can use these additional options on the nfs mount. This makes the operation fail if it fails after a specific number of retries.

# Details

soft:
Screenshot 2024-04-03 at 11 48 15

Basically, if we don't have soft, it will never "fail" and will just hang forever trying to reconnect.

**Note:** In terms of soft causing data corruption, I think if we say that "if file access fails once on a pod (tentacle or script pod) we will restart the tentacle deployment and fail the scripts in question" then we shouldn't have an issue with data corruption.

timeo:
Screenshot 2024-04-03 at 11 48 23

how long it waits before failing a connection in tenths of a second. The standard is 60 seconds.

retrans:
Screenshot 2024-04-03 at 11 48 29

how many times it will retry before failing.

# Testing

With the settings I have now (`soft, timeo=5, retrans=5`) I get the following output in manual testing:
Screenshot 2024-04-03 at 11 47 42

I've done more tests where I saw it take up to 11.5 seconds to fail too so the range is roughly 3-12 seconds for accessing a broken mount to fail. If you start the nfs pod back on the same node, the mount becomes valid again and returns the `ls` command instantly.

# Testing update

I've adjusted the numbers to `timeo=50, retrans=4`. This gives each connection time to recover (~5 seconds) and in total it takes between 25 and 50 seconds approximately.

Screenshot 2024-04-03 at 15 26 44

### DIFF
--- a/charts/kubernetes-agent/templates/nfs-pv.yaml
+++ b/charts/kubernetes-agent/templates/nfs-pv.yaml
@@ -16,8 +16,8 @@ spec:
     - nfsvers=4.1
     - lookupcache=none
     - soft
-    - timeo=5
-    - retrans=5
+    - timeo=50
+    - retrans=4
   csi:
     driver: nfs.csi.k8s.io
     volumeHandle: {{ printf "%s/octopus##" (include "nfs.serverAddress" .)}}
--- a/charts/kubernetes-agent/templates/nfs-pv.yaml
+++ b/charts/kubernetes-agent/templates/nfs-pv.yaml
@@ -16,7 +16,6 @@ spec:
     - nfsvers=4.1
     - lookupcache=none
     - soft
-    - bg
     - timeo=5
     - retrans=5
   csi:
--- a/charts/kubernetes-agent/templates/nfs-pv.yaml
+++ b/charts/kubernetes-agent/templates/nfs-pv.yaml
@@ -18,6 +18,7 @@ spec:
     - soft
     - bg
     - timeo=5
+    - retrans=5
   csi:
     driver: nfs.csi.k8s.io
     volumeHandle: {{ printf "%s/octopus##" (include "nfs.serverAddress" .)}}
--- a/charts/kubernetes-agent/templates/nfs-pv.yaml
+++ b/charts/kubernetes-agent/templates/nfs-pv.yaml
@@ -15,6 +15,9 @@ spec:
   mountOptions:
     - nfsvers=4.1
     - lookupcache=none
+    - soft
+    - bg
+    - timeo=5
   csi:
     driver: nfs.csi.k8s.io
     volumeHandle: {{ printf "%s/octopus##" (include "nfs.serverAddress" .)}}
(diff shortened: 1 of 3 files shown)


charts/kubernetes-agent/templates/nfs-pv.yaml
```diff
@@ -15,6 +15,9 @@ spec:
   mountOptions:
     - nfsvers=4.1
     - lookupcache=none
+    - soft
+    - bg
+    - timeo=5
```
APErebus, line 20: what do `soft` and `bg` do?

scme0: I took a look at the man page to see what these did so I could accurately relay that info and realised that timeo was in tenths of a second. So I started experimenting again and found that:
1. `bg` is not required
2. there seems to be a slightly better setup than what I had originally (using `retrans` and `timeo`)
3. `soft` is required

scme0: soft:
Screenshot 2024-04-03 at 11 48 15

Basically, if we don't have soft, it will never "fail" and will just hang forever trying to reconnect.

timeo:
Screenshot 2024-04-03 at 11 48 23

how long it waits before failing a connection in tenths of a second. The standard is 60 seconds.

retrans:
Screenshot 2024-04-03 at 11 48 29

how many times it will retry before failing.

With the settings I have now (`soft, timeo=5, retrans=5`) I get the following output in manual testing:
Screenshot 2024-04-03 at 11 47 42

I've done more tests where I saw it take up to 11.5 seconds to fail too so the range is roughly 3-12 seconds for accessing a broken mount to fail. If you restart the nfs pod, it returns instantly again.

scme0: In terms of soft causing data corruption, I think if we say that "if file access fails once on a pod (tentacle or script pod) we will restart the tentacle deployment and fail the scripts in question" then we shouldn't have an issue with data corruption.

charts/kubernetes-agent/templates/nfs-pv.yaml
```diff
@@ -15,6 +15,9 @@ spec:
   mountOptions:
     - nfsvers=4.1
     - lookupcache=none
+    - soft
+    - timeo=5
```
APErebus, line 19: this is 500ms?

scme0: yep - after speaking with @liam-mackie I'm going to change these values to `timeo=100` and `retrans=6` so it's approximately 1 minute of retries before a pod will fail.

APErebus (approved):
👍

PR #105, opened 2024-04-05
Read Pod logs from the K8s API

flin-8:
[sc-71685]

This PR updates the Tentacle version to start reading Pod logs from the K8s API. As part of this, we also need to set the appropriate permissions for Tentacle.


APErebus (approved):

scme0 (approved):
🚀

PR #108, opened 2024-04-08
Support providing test data to ConfigMap

APErebus:
In the Tentacle integration tests, we need a way to pre-provide config map data, so Tentacle doesn't try and register itself with a server that doesn't exist

### DIFF
--- a/charts/kubernetes-agent/values.yaml
+++ b/charts/kubernetes-agent/values.yaml
@@ -99,6 +99,7 @@ resources: {}
 #   memory: 128Mi
 
 
+# Used for integration testing to avoid registring with a real Octopus Server
 testing:
   tentacle:
     configMap:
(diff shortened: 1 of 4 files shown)


liam-mackie (approved):
LGTM! Do you need a changeset to use a versioned release in the tests?

charts/kubernetes-agent/values.yaml
```diff
@@ -97,3 +97,10 @@ resources: {}
 # requests:
 #   cpu: 100m
 #   memory: 128Mi
+
+
+# Used for integration testing to avoid registring with a real Octopus Server
+testing:
```
flin-8, line 103: Out of curiosity, what value does the test inject?

APErebus: https://github.com/OctopusDeploy/OctopusTentacle/pull/870/files?w=1#diff-e67e3ac4eb073effa4a495a05b2cf8f01b175807e075f3d88728b5353a0f3e21R88-R94

APErebus:
> LGTM! Do you need a changeset to use a versioned release in the tests?

Yes. Thanks for the reminder :)

PR #109, opened 2024-04-11
Support multiple polling connections with a default of 5

APErebus:
Tentacle `8.1.1332` supports opening multiple polling connections, add a new value to set this new environment variable.

The current default is `5`, but that was pulled out of thin air, so happy to change

### DIFF
--- a/charts/kubernetes-agent/package.json
+++ b/charts/kubernetes-agent/package.json
@@ -3,5 +3,11 @@
   "version": "0.7.1",
   "private": true,
   "description": "The Octopus Kubernetes Agent",
-  "author": "Octopus Deploy Ptd Ltd"
+  "author": "Octopus Deploy Ptd Ltd",
+  "scripts": {
+    "test": "cross-env-shell docker run -ti --rm -v $INIT_CWD:/apps helmunittest/helm-unittest:latest ."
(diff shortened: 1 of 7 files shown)


scme0 (approved):

charts/kubernetes-agent/package.json
```diff
@@ -3,5 +3,11 @@
   "version": "0.7.1",
   "private": true,
   "description": "The Octopus Kubernetes Agent",
-  "author": "Octopus Deploy Ptd Ltd"
+  "author": "Octopus Deploy Ptd Ltd",
+  "scripts": {
+    "test": "cross-env-shell docker run -ti --rm -v $INIT_CWD:/apps helmunittest/helm-unittest:latest ."
```
APErebus, line 8: I can't run the unittest code in windows, so this just runs it in a docker container. just has to be executed via `npm test`

charts/kubernetes-agent/tests/tentacle-container-env-vars_test.yaml
APErebus, line 1: new test class for checking env vars are correctly set

APErebus:
@scme0 I added a unit test

PR #110, opened 2024-04-11
Add NFS Watchdog to helm chart

liam-mackie:
This PR adds the NFS Watchdog to the helm chart, along with associated tests. This also makes it so that the cluster rolebinding for pods is added to the release namespace by default when NFS is enabled, as we need DELETE access to pods in this namespace for the NFS watchdog to function.

### DIFF
--- a/charts/kubernetes-agent/templates/pod-rolebinding.yaml
+++ b/charts/kubernetes-agent/templates/pod-rolebinding.yaml
@@ -18,3 +18,17 @@ roleRef:
   name: {{ $podClusterRoleName }}
   apiGroup: rbac.authorization.k8s.io
 {{- end -}}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ printf "%s-binding" $podServiceAccountFullName }}
+  namespace: {{ .Release.Namespace | quote }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ $podServiceAccountName }}
+    namespace: {{ $.Release.Namespace | quote }}
+roleRef:
+  kind: ClusterRole
+  name: {{ $podClusterRoleName }}
+  apiGroup: rbac.authorization.k8s.io
--- a/charts/kubernetes-agent/templates/pod-rolebinding.yaml
+++ b/charts/kubernetes-agent/templates/pod-rolebinding.yaml
@@ -18,7 +18,6 @@ roleRef:
   name: {{ $podClusterRoleName }}
   apiGroup: rbac.authorization.k8s.io
 {{- end -}}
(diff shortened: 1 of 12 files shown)


charts/kubernetes-agent/templates/pod-rolebinding.yaml
```diff
@@ -17,4 +17,20 @@ roleRef:
...
+  name: {{ printf "%s-binding" $podServiceAccountFullName }}
+  namespace: {{ .Release.Namespace | quote }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ $podServiceAccountName }}
+    namespace: {{ $.Release.Namespace | quote }}
+roleRef:
+  kind: ClusterRole
+  name: {{ $podClusterRoleName }}
+  apiGroup: rbac.authorization.k8s.io
```
APErebus, line 35: 💭 Should we have a Role that is _just_ for the script pod that has only pod delete permissions 🤔
Is there a way we can stop if from being able to kill pods that aren't script pods?

liam-mackie: We definitely could have a specific role with only delete - we will need it to be at least namespace scoped, as that's the smallest scope 🙁

APErebus: Yeah, let's do that. Then we don't give the script pod itself too many permissions to the namespace of the agent

liam-mackie: Done in [c59228f](https://github.com/OctopusDeploy/helm-charts/pull/110/commits/c59228f19c8575e81f442e93c4b0e3ab8062dc80) :)

charts/kubernetes-agent/tests/pod-clusterole_test.yaml
```diff
@@ -12,13 +12,22 @@ tests:
...
-        resources: ["pods"]
-        verbs: ["get", "watch", "list"]
+    - contains:
+        path: rules
+        content:
+          apiGroups:
+          - '*'
+          resources: ["pods"]
+          verbs: ["get", "watch", "list"]
+      documentIndex: 0
```
liam-mackie, line 22: 0 is the one we're testing

charts/kubernetes-agent/tests/pod-clusterole_test.yaml
```diff
@@ -12,13 +12,22 @@ tests:
...
+    - contains: 
+        path: rules
+        content:
+          apiGroups:
+          - '*'
+          resources: ["pods"]
+          verbs: ["delete", "list"]
+      documentIndex: 1
```
liam-mackie, line 30: 1 is the newly created delete-only role!

APErebus (approved):
🚀 Awesome stuff @liam-mackie

PR #115, opened 2024-04-17
Remove ability to change replica count

APErebus:
We don't currently support multiple replicas of the Tentacle container, so let's remove the option to change it


flin-8 (approved):
LGTM

PR #119, opened 2024-04-22
Increase default nfs volume and claim size

tleed5:
# Background
[SC-73899]
When the NFS pod runs out of space it will restart and causes all of the data in the `/octopus` directory to be lost

# Result 
- Increased the default size of the `emptyDir` volume used by the NFS pod to `10Gi`
- Increased the default size of the PVC used by the tentacle pod to `10Gi`


APErebus (approved):
Awesome! Thanks

PR #124, opened 2024-04-24
Adding a suffix for the nfs size to the pvc and pv to allow size upgrades

scme0:
This allows us to increase the default in the future and upgrades will work as expected.


APErebus (approved):
Excellent solution 👏

scme0:
> Excellent solution 👏

All credit goes to @liam-mackie for this one! 🙇

PR #125, opened 2024-04-24
Space checks for tentacle

liam-mackie:
When running Octopus Tentacle in the Kubernetes Agent, we can't reliably get disk space on the mounted volume using native disk space tools. This is because in some cases (especially using the default NFS implementation), the volume is actually an NFS export of a folder on the node. This means that disk space tools (like `df`) show the disk space free on the _node_ rather than the disk space that we've configured in Kubernetes to be available. When we exceed the disk space we've configured in Kubernetes, `kubelet` will evict the NFS pod.

This PR adds an environment variable to support [changes in Tentacle](https://github.com/OctopusDeploy/OctopusTentacle/pull/904) to use `du` to check free disk space.

In addition, this pulls the logging back from `debug` to `info`, since there's some extra logging in the Tentacle changes, which we don't want to mess around with much.

### DIFF
--- a/.changeset/unlucky-days-breathe.md
+++ b/.changeset/unlucky-days-breathe.md
@@ -1,0 +1,5 @@
+---
+"kubernetes-agent": minor
+---
+
+Update to add required PERSISTENTVOLUMESIZE variable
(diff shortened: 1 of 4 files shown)


APErebus (approved):
When the Tentacle PR merges, update this PR with the Tentacle build version so it all goes in together :)

APErebus:
Also, can you link the Tentacle PR in this PR so we have a record of the relationship

PR #130, opened 2024-04-24
Reorganise values.yaml in preparation for 1.0.0 release

APErebus:
In preparation for the `1.0.0` release of the Helm Chart, we have done a reorganization of the `values.yaml` file to be more inline with typical design.

The main changes are `tentacle` becomes `agent` and we've re-organized under the `scriptPods` section

I also changed the name of the script pod roles & rolebindings to be `octopus-agent-scripts` vs `octopus-agent-pod`

Also removed listening tentacle service & configuration as it's not supported.

I chose not to move the files around just yet, mainly to avoid git from doing weird diff things


APErebus:
I have deployed this locally with an NFS pod and the Server PR with the updated helm command and it all worked.
I need to do a test with a PV, but Kind doesn't have a ReadWriteMany storage provider

APErebus:
I ran this using an `azurefile` storage class in AKS and it correctly did not add NFS resources and ran. Marking this as ready for review as the 2 primary modes via the UI work

charts/kubernetes-agent/templates/pod-clusterroles.yaml
```diff
@@ -17,12 +17,12 @@ rules:
   verbs:
   - '*'
 {{- end }}
-{{- if and .Values.persistence.nfs.watchdog.enabled .Values.persistence.nfs.enabled}}
+{{- if and .Values.persistence.nfs.watchdog.enabled (not .Values.persistence.storageClassName)}}
```
APErebus, line 20: I couldn't make this into a helper due to templating issues. We do this check in a number of places, so would be nice to make common

scme0 (approved):
We paired on this review and it all looks good. Great to have the additional snapshot tests for extra assurance! Nice work! 🎉

PR #133, opened 2024-04-25
Add snapshot tests for all resource files

APErebus:
To provide extra verification that the changes in #130 have no untended side-affects for the default options, I've added snapshot tests for all resource files


scme0 (approved):
🚀

PR #142, opened 2024-04-30
Add machine policy option to values

scme0:
Allow users to specify a custom machine policy

[[sc-77359]](https://app.shortcut.com/octopusdeploy/story/77359/allow-users-to-set-the-machine-policy-in-the-helm-chart)

### DIFF
--- a/charts/kubernetes-agent/templates/tentacle-deployment.yaml
+++ b/charts/kubernetes-agent/templates/tentacle-deployment.yaml
@@ -46,6 +46,10 @@ spec:
               value: {{ join "," .Values.agent.targetEnvironments | quote }}
             - name: "TargetRole"
               value: {{ join "," .Values.agent.targetRoles | quote }}
+            {{- with .Values.agent.machinePolicy }}
+            - name: "MachinePolicy"
+              value: {{ . | quote }}
+            {{- end }}
             - name: "DefaultNamespace"
               value: {{ .Values.agent.defaultNamespace | quote }}
             - name: "OCTOPUS__K8STENTACLE__NAMESPACE"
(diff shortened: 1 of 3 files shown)


charts/kubernetes-agent/templates/tentacle-deployment.yaml
```diff
@@ -46,6 +46,10 @@ spec:
               value: {{ join "," .Values.agent.targetEnvironments | quote }}
             - name: "TargetRole"
               value: {{ join "," .Values.agent.targetRoles | quote }}
+            {{- with .Values.agent.machinePolicy }}
+            - name: "MachinePolicy"
+              value: {{ . | quote }}
+            {{- end }}
```
APErebus, line 52: Is this env var already supported in the docker image?

charts/kubernetes-agent/values.yaml
```diff
@@ -13,6 +13,7 @@ agent:
   space: "Default"
   targetEnvironments: []
   targetRoles: []
+  machinePolicy: ""
```
APErebus, line 16: would `machinePolicyName` be more accurate/clearer?

charts/kubernetes-agent/templates/tentacle-deployment.yaml
scme0: Yep 👍 Tested and it works with just this change

charts/kubernetes-agent/values.yaml
scme0: I wasn't sure because Environments and Space also match by name but we don't specify it 🤔

APErebus: I think let's be specific if we can

scme0: Done 👍

APErebus (approved):

PR #145, opened 2024-05-03
Add initial c# test structure for kubernetes-agent

liam-mackie:
Simple scaffolding to allow us to start running xunit integration tests for the `kubernetes-agent` helm chart


APErebus (approved):
I just tweaked the solution & project name to be more C#-like and updated the build config

PR #147, opened 2024-05-07
Version Kubernetes Agent Chart

Octobob:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and publish to npm yourself or [setup this action to publish automatically](https://github.com/changesets/action#with-publishing). If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Patch Changes

-   578d0c7: Update Tentacle to 8.1.1588. Introduces new `KubernetesScriptServiceV1`.

PR #148, opened 2024-05-09
Allow BYO subscription ID and tentacle certificate

tleed5:
# Background

[SC-78187]
Related to https://github.com/OctopusDeploy/OctopusTentacle/pull/927 and https://github.com/OctopusDeployLabs/terraform-provider-octopusdeploy/pull/636

Includes these tentacle changes https://github.com/OctopusDeploy/OctopusTentacle/compare/8.1.1588...8.1.1617

# Results
Add new values for passing through subscription id and certificate
Two new values for subscriptionID and certificate
New secret to store the certificate


charts/kubernetes-agent/templates/tentacle-certificate-secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
APErebus, line 1: We should only create this secret if the certificate is set

charts/kubernetes-agent/tests/tentacle-deployment_test.yaml
```diff
@@ -129,3 +129,23 @@ tests:
...
+
+  - it: "doesn't include certificate when not set"
+    set:
+      agent:
+        certificate: ""
+    asserts:
+      - notContains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: "TentacleCertificateBase64"
```
APErebus, line 151: Can you move these into the `tentacle-container-env-vars_test.yaml` please?

charts/kubernetes-agent/README.md
```diff
@@ -1,7 +1,6 @@
 # kubernetes-agent
 
-
-![Version: 1.0.3](https://img.shields.io/badge/Version-1.0.3-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 8.1.1588](https://img.shields.io/badge/AppVersion-8.1.1588-informational?style=flat-square) 
+![Version: 1.0.4](https://img.shields.io/badge/Version-1.0.4-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 8.1.1588](https://img.shields.io/badge/AppVersion-8.1.1588-informational?style=flat-square)
```
APErebus, line 3: You may want to update to `main` because this should already by `1.0.4`

APErebus (approved):
You'll also need to update the Tentacle version before merging this

charts/kubernetes-agent/README.md
tleed5: Currently in `main` the readme still says `1.0.3` https://github.com/OctopusDeploy/helm-charts/blob/640dc033cd36dea75f88f0103206a1595a703c6b/charts/kubernetes-agent/README.md?plain=1#L4 

I think it's because the versioning bump happens in a separate PR to when the readme is generated e.g. https://github.com/OctopusDeploy/helm-charts/pull/147 didn't regenerate the readme after changing the version numbers

tleed5: This should be fixed with https://github.com/OctopusDeploy/helm-charts/pull/150 so going forward it should update the readme on versioning PRs

APErebus (approved):
Awesome stuff. Thanks @tleed5

PR #153, opened 2024-05-13
Update Kubernetes Agent readme with Octopus Deploy version

APErebus:
Just to give more context about what version of Octopus Server supports this version of the Agent


charts/kubernetes-agent/README.md.gotmpl
```diff
@@ -1,13 +1,16 @@
...
 
 {{ template "chart.maintainersSection" . }}
 
 {{ template "chart.sourcesSection" . }}
 
-{{ template "chart.valuesSection" . }}
\ No newline at end of file
+{{ template "chart.valuesSection" . }}
+
+{{ template "helm-docs.versionFooter" . }}
```
liam-mackie, line 16: Is there a reason we're adding the `helm-docs` footer in?

APErebus: Yeah. I think it's appropriate that we acknowledge the use of this tool via the footer. Give readers information that they might want to use the tool as well

liam-mackie: The only reason I didn't was to stop the version from showing diffs when the container updates on a dev machine vs. CI, not to hide the fact we use the tool 😅

APErebus: oo, fair... I'll remove it this and replace with with a fixed footer that won't change

liam-mackie (approved):
LGTM! Thanks for the cleanup!

PR #154, opened 2024-05-22
Add metrics configmap entry to helm chart

rain-on:
This config map will be used by the agent to track failures (or other metrics).

Context: The health-check wrapper script will scrape this config map, and post the content back to OctopusDeploy a Json-blob in an OctopusVariable. 

This does not include options to populate from tests, its a first pass to get data being logged.

### DIFF
--- a/.changeset/cuddly-geckos-build.md
+++ b/.changeset/cuddly-geckos-build.md
@@ -1,0 +1,5 @@
+---
+"kubernetes-agent": minor
+---
+
+Add ConfigMap for agent metrics persistence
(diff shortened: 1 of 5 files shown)


scme0 (commented):
You need to add a changeset to this PR so that it versions correctly:
1. Install the changeset tool: `npm install -g changeset`
2. In the helm-charts repo root directory, run `pnpm changeset`

scme0:
Maybe for completeness, add a snapshot test? I know that it's just a blank configmap right now but just in case 🤷‍♂️ 

Probably just something like this would be fine: charts/kubernetes-agent/tests/tentacle-configmap_test.yaml

scme0 (approved):
Nice!

PR #155, opened 2024-05-23
Update agent github actions to always run (jobs are filtered to skip if no changes to agent code)

scme0:
# Background
We want to make sure the agent github actions are `required` but we need to make sure they run for all PRs (even if the jobs themselves get skipped). This is so that PRs with no changes to the agent helm chart will not be blocked from merging.

Notes from Github docs relating to this:
Screenshot 2024-05-23 at 10 56 32

# Results

Using a solution from here: https://stackoverflow.com/a/70839666

I moved the test step into the main workflow so I could make a dependency chain with tests running first This also means we only have to add the path_filter style action once.


scme0:
No Changeset required as we aren't making changes to the chart itself.

APErebus (approved):

PR #161, opened 2024-05-24
(WIP) Additional Isolated Servers support

scme0:
If we want to support an agent registering itself with multiple separate instances of server, we need to define all the information for each server instance/cluster.

My suggestion is that we add an array property where users can define this data for each additional server they want to register. This can be passed into the Kubernetes tentacle environment variable as a JSON blob which is pulled apart by the configure script and multiple calls to `register-k8s-cluster` can be made.


scme0:
Closing as out of scope.

PR #163, opened 2024-05-27
Remove branch filter for PR requests

APErebus:
We don't need as we are doing all branches


scme0 (approved):
Awesome, thanks for this!

PR #165, opened 2024-05-28
Add token to checkout action

scme0:
The Versioning workflow creates a PR but the Kubernetes build/test/publish workflow doesn't run when the PR is created.

We think this is because it's using the `GITHUB_TOKEN` to authenticate.

I believe updating the checkout action to use the changeset token should make all subsequent usages of git use the same token which should fix the problem. Hard to say if it will work until after it's merged though.


scme0:
No changeset required because there is no change to the helm chart.

APErebus (approved):

PR #166, opened 2024-05-30
Update tentacle version to add ability to update ServerCommsAddress with `helm upgrade` command

scme0:


.changeset/eleven-walls-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"kubernetes-agent": patch
```
APErebus, line 2: minor as it's a new feature :)

APErebus (approved):

PR #167, opened 2024-05-30
Add startup probe for tentacle container

scme0:
# Background

Currently you can run the helm upgrade command for a Kubernetes agent and it will install and helm will be happy even if there is no server running to connect to.

When the tentacle starts up, Kubernetes sees the container running and helm just thinks it's good to go, even though the configuration/initialisation hasn't completed yet. This can put the agent in a situation where the Tentacle container is constantly bootlooping, unable to do anything while the helm chart returned happy.

# Results

This change is partnered with a Tentacle change (https://github.com/OctopusDeploy/OctopusTentacle/pull/953) which creates an empty file when the tentacle container is finished initialising on startup. The startup probe simply checks to see if the file exists and as soon as it does the container is considered "ready" and then the helm chart can complete successfully. 

[sc-80003]

### DIFF
--- a/.changeset/neat-tigers-turn.md
+++ b/.changeset/neat-tigers-turn.md
@@ -1,0 +1,5 @@
+---
+"kubernetes-agent": minor
+---
+
+Add startupProbe to Kubernetes tentacle container to ensure helm upgrade command doesn't return successful when the tentacle failed to initialise.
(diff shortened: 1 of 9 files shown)


scme0:
⚠️ This is a draft because I need to update the Tentacle version with the change from https://github.com/OctopusDeploy/OctopusTentacle/pull/953 before merging this.

APErebus (approved):
Nice work mate :)

PR #171, opened 2024-06-01
Add Tenant and Tenant tag support

scme0:
Quite a simple change which adds properties which are passed through to the Kubernetes tentacle.

I'm not actually quite sure what a Tenant Tag is? I've added it because it was also in the `configure-and-run.sh` in Tentacle but we could remove it if it's not something people would use?

[sc-80799]


charts/kubernetes-agent/values.yaml
```diff
@@ -33,16 +32,25 @@ agent:
...
   # @section -- Agent values
   space: "Default"
   # -- The target environments to register the agent with
   # @section -- Agent values
-  targetEnvironments: []  
+  targetEnvironments: []
   # -- The target roles to register the agent with
   # @section -- Agent values
   targetRoles: []
+  # -- Can be "Untenanted", "TenantedOrUntenanted" or "Tenanted".
```
APErebus, line 45:
```suggestion
  # -- Can be `Untenanted`, `TenantedOrUntenanted` or `Tenanted`.
```

APErebus (approved):
Will also need the Tentacle container updated as well

PR #172, opened 2024-06-03
Update Kubernetes agent tentacle container repository name and bump version

scme0:
Now that Tentacle has been updated to publish the Kubernetes agent Tentacle container as `kubernetes-agent-tentacle` this PR needs to be merge before others so that we can continue to bump the Tentacle container version. Otherwise the Integration tests will fail because helm won't be able to pull the image (as it will be pulling the `kubernetes-tentacle` image).

[sc-80191]


APErebus (approved):
Awesome!

PR #173, opened 2024-06-04
Add pods/exec permission to tentacle deployment

scme0:
This is required if we want to be able to use the `Client.NamespacedPodExecAsync()` command like in:
https://github.com/OctopusDeploy/OctopusTentacle/pull/960

https://github.com/OctopusDeploy/OctopusDeploy/pull/25256


scme0:
closed as no longer necessary

PR #180, opened 2024-06-11
Use Artifactory to source the Kubernetes agent helm chart instead of Dockerhub

scme0:
# Background

We've been hitting rate limiting because docker is tightening up their rate limits.

# Results

Instead of doing what we do in server which is to return inconclusive when dockerhub rate limits our request, we are just going to use Artifactory. This will eliminate the rate limiting but could cause failures if artifactory is done. Because this test doesn't have to run on the same scale as server, we don't need to stress too much about this. If it does fail, we can just rerun the test.


scme0:
☝️ No changeset required as this is just changes to tests

liam-mackie (approved):
LGTM! Just checking, does 1.*.* resolve prerelease versions or just main versions?

scme0:
> LGTM! Just checking, does 1.*.* resolve prerelease versions or just main versions?

As discussed in slack, we've tested this and `1.*.*` does not resolve prerelease versions. `1.*.*-0` is what to use if you want to resolve the most recent pre-release version 👍

PR #184, opened 2024-06-19
Helm v2 chart to support registering the Kubernetes Agent as a Worker

kevjt:
Draft proposal of what the new Helm chart should look like to allow the agent to be registered as an Octopus Worker.


charts/kubernetes-agent/values.yaml
```diff
@@ -60,16 +45,40 @@ agent:
...
   # -- The number of polling TCP connections to open with the target Octopus Server
   # @section -- Agent values
   pollingConnectionCount: 5
-
   # -- True if events should be scraped and added to the metrics config map
   # @section -- Agent values
   enableMetricsCapture: true
+  # -- Can be `DeploymentTarget` or `Worker`.
+  # @section -- Agent values
+  registerAs: ""
```
kevjt, line 56: I made this accept a string instead of an array of string because at the moment we would like to only allow registering as **either a worker or deployment target** but not both.

Still, I am keen on getting feedback about whether this is the right thing to do, because there is a question of whether we will end up allowing the option to register the agent as both at the same time.

If we are likely to move towards allowing both, then another option (that I prefer) is to replace this field with fields called something like `isRegistered` on child types (i.e. `deploymentTarget.isRegistered` and `worker.isRegistered`).

charts/kubernetes-agent/templates/tentacle-deployment.yaml
```diff
@@ -51,22 +51,26 @@ spec:
...
             - name: "TargetTenant"
-              value: {{ join "," .Values.agent.targetTenants | quote }}
+              value: {{ join "," .Values.agent.deploymentTarget.tenants | quote }}
             - name: "TargetTenantTag"
-              value: {{ join "," .Values.agent.targetTenantTags | quote }}
+              value: {{ join "," .Values.agent.deploymentTarget.tenantTags | quote }}
             - name: "TargetTenantedDeploymentParticipation"
-              value: {{ .Values.agent.targetTenantedDeploymentParticipation | quote }}
+              value: {{ .Values.agent.deploymentTarget.tenantedDeploymentParticipation | quote }}
+            - name: "WorkerPoolIds"
```
kevjt, line 66: I figured that there is no harm in passing through these environment variables all the time regardless of the selected registration option. Since the `configure-and-run.sh` script on Tentacle will need to respond to an explicit selector field anyway, and simply passing through here removes the need for another conditional block in the helm template.

charts/kubernetes-agent/values.yaml
```diff
@@ -9,12 +9,12 @@ imagePullSecrets: []
 #   - name: "registry-secret-name"
 
 agent:
+  # -- The name of the agent
+  # @section -- Agent values
+  name: ""
```
kevjt, line 14: The current thinking is that the agent will have a single name field which will be used for the Octopus Server name of either the worker or the target. This is also what gets used to determine the kubernetes namespace etc. This means that should the user want to install an agent worker using the same name as an existing agent target, then there will be a namespace conflict in their cluster we simply give a helm command the way things are at the moment.

Would like to hear some thoughts about this too. There are several different ways this problem can be solved..

charts/kubernetes-agent/templates/tentacle-deployment.yaml
```diff
@@ -42,7 +42,7 @@ spec:
             - name: "ACCEPT_EULA"
               value: {{ .Values.agent.acceptEula | quote }}
             - name: "TargetName"
-              value: {{ .Values.agent.targetName | quote }}
+              value: {{ .Values.agent.name | quote }}
             - name: "ServerCommsAddresses"
```
kevjt, line 46: I was contemplating whether or not to put ServerCommsAddresses under its own category namespace, and everything else under another. Something like `agent.ServerCommsAddresses` and everything else under `agent.initial._...`
The idea is to provide some distinction between "only-on-install" options that can't be updated in Tentacle after the first time install vs updatable fields. However, the extra namespace feels a bit overkill and can add confusion.

Maybe it's best to simply be clear with a comment or two in the docs.

charts/kubernetes-agent/templates/tentacle-deployment.yaml
```diff
@@ -42,7 +42,7 @@ spec:
...
               value: {{ join "," .Values.agent.serverCommsAddresses | quote }}
             - name: "ServerCommsAddress"
```
kevjt, line 48: This can be removed as part of the major bump.

@scme0 please let me know if there are other fields that can be cleaned up.

scme0: The two that come to mind are:
- `ServerCommsAddress` can be removed (we'll just use `ServerCommsAddresses` all the time instead)
- `targetRoles` should use the word `tags` instead of roles.

scme0: I'll create a server PR to change the UI to always use `ServerCommsAddresses`, even if you only have a single value.

scme0: @APErebus Are there any other properties that we should change in v2?

charts/kubernetes-agent/README.md
```diff
@@ -27,28 +27,40 @@ A Helm chart for the Octopus Kubernetes Agent
...
 | agent.debug.disableAutoPodCleanup | bool | `false` | Disables automatic pod cleanup |
-| agent.defaultNamespace | string | `""` | The default Kubernetes namespace for deployments |
 | agent.enableMetricsCapture | bool | `true` | True if events should be scraped and added to the metrics config map |
 | agent.image | object | `{"pullPolicy":"IfNotPresent","repository":"octopusdeploy/kubernetes-agent-tentacle","tag":"8.1.1758"}` | The repository, pullPolicy & tag to use for the agent image |
 | agent.logLevel | string | `"Info"` | The log level of the agent. Logs are written to the pod logs as well as to file |
 | agent.machinePolicyName | string | `""` | The machine policy to register the agent with |
 | agent.metadata | object | `{"annotations":{},"labels":{}}` | Additional metadata to add to the agent pod & container |
+| agent.name | string | `""` | The name of the agent |
 | agent.pollingConnectionCount | int | `5` | The number of polling TCP connections to open with the target Octopus Server |
+| agent.registerAs | string | `""` | Can be `DeploymentTarget` or `Worker`. |
```
APErebus, line 37: This I'm not sure about, especially if we end up offering the ability to register as both 🤔

kevjt: Yea that's what I was not sure about https://github.com/OctopusDeploy/helm-charts/pull/184#discussion_r1645396933

The challenge of having a permissive structure now is how we convey that it can't be both. Perhaps with extra validation on the tentacle `configure-and-run.sh` script side?

kevjt: Actually the more I think about it.. we should leave the option open to have both in the Helm Chart, and it's probably sufficient to have validation on the Tentacle start script which will return an error message for now if both are set.

kevjt: In which case I prefer to have an `.isRegistered` option on each of the worker and deploymentTarget levels.

scme0: There is kinda prior art for having a property with "option 1", "option 2", "option 1 or option 2" in the `tenantedDeploymentParticipation` property. But in our case it would be "option 1 AND option2" I guess. 🤔

scme0: The one thing I don't like about the `.isRegistered` property is it does sound like you could change it later and it will update the registration. Maybe `.registerOnInitialisation` or `.registerAsWorker`.

charts/kubernetes-agent/values.yaml
```diff
@@ -60,16 +45,45 @@ agent:
...
+      defaultNamespace: ""
+    
+  worker:
+    # -- Set to register the agent as a Worker using the provided initial values
+    # @section -- Agent as Worker values
+    enabled: false
+    initial:
+      # -- The worker pool IDs to associate with the worker
+      # @section -- Agent as Worker initial values
+      workerPoolIds: [ ]
```
scme0, line 86: Lets go with` workerPools` I think... (without `Ids`)

kevjt: yes good point. Since we are allowing names ids or slugs

PR #186, opened 2024-06-20
Allow configuring proxy environment variables

hnrkndrssn:
A customer is working on PoC running Octopus in Kubernetes but there's no way to configure the proxy to use, this PR adds the ability to configure the `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, and `NO_PROXY` environment variables in the container.

[sc-82304]

### DIFF
--- a/charts/octopus-deploy/tests/statefulset_test.yaml
+++ b/charts/octopus-deploy/tests/statefulset_test.yaml
@@ -78,15 +78,19 @@ tests:
       - ./values/required.yaml
       - ./values/proxy.yaml
     asserts:
-      - contains:
-          path: spec.template.spec.containers[0].env
-          content:
-            name: HTTP_PROXY
-            value: http://myproxy.com
+      - equal:
+          path: spec.template.spec.containers[0].env[?(@.name == 'HTTP_PROXY')].value
+          value: http://myproxy.com
         documentIndex: 0
-      - contains:
-          path: spec.template.spec.containers[0].env
-          content:
-            name: HTTPS_PROXY
-            value: https://myproxy.com
+      - equal:
+          path: spec.template.spec.containers[0].env[?(@.name == 'HTTPS_PROXY')].value
+          value: https://myproxy.com
+        documentIndex: 0
+      - equal:
+          path: spec.template.spec.containers[0].env[?(@.name == 'ALL_PROXY')].value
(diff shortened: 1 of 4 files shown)


charts/octopus-deploy/values.yaml
```diff
@@ -104,6 +104,16 @@ octopus:
...
 
+  proxy:
+    # the proxy to use for http requests
+    http:
+    # the proxy to use for https requests
+    https:
+    # the proxy to use for http and/or https requests, in case the above two values are not defined
+    all:
+    # a comma-separated list of hostnames that should be excluded from proxying
+    none:
```
APErebus, line 115: It would be more "correct" if this was an array type, that then you can do a join in when using it. See https://github.com/OctopusDeploy/helm-charts/blob/main/charts/kubernetes-agent/templates/tentacle-deployment.yaml#L46-L47

An example 

Also can we rename this to be clearer?

```suggestion
    excludedHostnames: []
```

charts/octopus-deploy/tests/statefulset_test.yaml
```diff
@@ -71,4 +71,22 @@ tests:
           content: 
             name: "ACCEPT_EULA"
             value: "N"
+        documentIndex: 0
+  
+  - it: proxy is configurable
+    values:
+      - ./values/required.yaml
+      - ./values/proxy.yaml
```
APErebus, line 79: Can we make sure we test all possible inputs, especially if we end up using an array for the excluded hostnames

MJRichardson (approved):
LGTM

charts/octopus-deploy/values.yaml
```diff
@@ -104,6 +104,16 @@ octopus:
...
     labels: {}
 
+  proxy:
+    # the proxy to use for http requests
+    http:
+    # the proxy to use for https requests
+    https:
+    # the proxy to use for http and/or https requests, in case the above two values are not defined
+    all:
+    # a comma-separated list of hostnames that should be excluded from proxying
```
APErebus, line 114: nit
```suggestion
    # an array of hostnames that should be excluded from proxying
```

hnrkndrssn: good catch!

PR #188, opened 2024-06-21
Add support for custom Octopus Server SSL/TLS certificates

APErebus:
Customers have reported issues where the agent cannot register due to the target Octopus Server having a custom or self-signed TLS certificate.

This PR adds support for this by adding a new value option.
```yaml
agent:
  serverCertificate: ""
```
When specified with an base64 encoded certificate, this mounts the certificate into `/etc/ssl/certs` using a configmap mount.

To support this, we also need the `kubernetes-agent-tentacle` docker container updated to perform a `openssl rehash /etc/ssl/certs` as the loaded certificate needs hashing before it can be used. This is coming in another PR: https://github.com/OctopusDeploy/OctopusTentacle/pull/967

This PR will also need the tentacle version updated when the above PR is merged

### DIFF
--- a/.changeset/slow-hotels-do.md
+++ b/.changeset/slow-hotels-do.md
@@ -1,0 +1,5 @@
+---
+"kubernetes-agent": minor
+---
+
+Add support for custom Octopus Server SSL/TLS certificates"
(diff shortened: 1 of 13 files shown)


liam-mackie (approved):
LGTM!

PR #192, opened 2024-06-21
replace _ with - in branch name for version in pre-release branches.

rain-on:
I keep making branches with underscores, which results in illegal version strings during GHA.

this updates the GHA to replace underscores with dashes in the version - yay.

### DIFF
--- a/.changeset/slimy-snails-sin.md
+++ b/.changeset/slimy-snails-sin.md
@@ -1,0 +1,5 @@
+---
+"kubernetes-agent": patch
+---
+
+Quality of life change to GHA workflow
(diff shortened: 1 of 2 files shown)


.changeset/slimy-snails-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+"kubernetes-agent": patch
+---
+
+Quality of life change to GHA workflow
```
scme0, line 5: Be more specific with this I think.

scme0 (approved):
Seems like a good idea 🤷‍♂️

.changeset/slimy-snails-sin.md
liam-mackie, line 1: Do you need a changeset since you're not changing code, just CI?

rain-on: I would say "no", there shouldn't be a changeset - but I wasn't sure how to prevent the check :/
Seems innocuous enough, but should work this out in future.

PR #194, opened 2024-06-24
Remove redundant jobs from Kubernetes agent GHA

scme0:
# Background

These were required because we had required checks on them but now they can be removed.

[sc-82969]


scme0:
☝️ No changeset required as we're just adjusting the GHAs

kevjt (approved):
lgtm as I can see these are clearly no-ops. Out of interest, where did the required checks come from?

scme0:
@kevjt Required checks come from repository configuration which is managed via https://github.com/OctopusDeploy/GithubBranchProtections

I adjusted the required checks here: https://github.com/OctopusDeploy/GithubBranchProtections/pull/752
and I added branch protections for release branches here: https://github.com/OctopusDeploy/GithubBranchProtections/pull/753

PR #195, opened 2024-06-24
Add release/kubernetes-agent/v* as a branch for GHAs

scme0:
# Background

Now that we support release branches, we need the GHAs to trigger on pushes to them.

# Results

We just need to add these as branches to the GHAs.

[sc-82968]


scme0:
☝️ Changeset not required as we're just updating the GHAs

kevjt (approved):

PR #196, opened 2024-06-24
Helm v2 pre-release chart to support registering the Kubernetes Agent as a Worker 

kevjt:
Similar to https://github.com/OctopusDeploy/helm-charts/pull/184

However, this is meant as a pre-release and so doesn't contain a lot of the variable renaming (we'll do things iteratively in the pre-release version).

[sc-82021]

### Associated PRs
- [Octopus.Client](https://github.com/OctopusDeploy/OctopusClients/pull/858)
- [Tentacle](https://github.com/OctopusDeploy/OctopusTentacle/pull/969)


scme0 (commented):
This is looking good! I think you just need to regenerate the docs.

charts/kubernetes-agent/README.md
```diff
@@ -42,14 +42,37 @@ A Helm chart for the Octopus Kubernetes Agent
...
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| agent.worker.enabled | bool | `false` | Set to register the agent as a Worker using the provided initial values |
+
+### Agent as Worker initial values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| agent.worker.initial.workerPools | list | `[]` | The worker pools to associate with the worker |
```
rain-on, line 75: do we need to mention that its a list of _strings_ - or is that implicit?

rain-on (approved):
I am by no means a helm expert - but it looks ok :)

scme0 (approved):
I think this is a nice starting point, We can iterate on it with what ever small fixes are required.

charts/kubernetes-agent/README.md
kevjt: I believe that is implicit. The doc generator tool produces this based on the default value in the values.yaml. The default there is simply `[]` which in helm/yaml terms can mean any type.

charts/kubernetes-agent/tests/__snapshot__/tentacle-deployment_test.yaml.snap
```diff
@@ -94,7 +100,7 @@ should match snapshot:
                   value: '{"requests":{"cpu":"25m","memory":"100Mi"}}'
                 - name: OCTOPUS__K8STENTACLE__PERSISTENTVOLUMESIZE
                   value: 10Gi
-              image: octopusdeploy/kubernetes-agent-tentacle:8.1.1908
+              image: docker.packages.octopushq.com/octopusdeploy/kubernetes-agent-tentacle:8.1.1925
```
kevjt, line 103: We will need to change this back to dockerhub when exiting pre-release mode

PR #197, opened 2024-06-24
Publish tag when a versioning commit merges

scme0:
# Background

We want to publish a tag with the version of the Kubernetes agent each time a Versioning PR is merged.

# Results

The versioning github action document will run a publish after the Versioning PR is merged to main.

Note: I can't really test this until we have a new Version to publish...

### DIFF
--- a/.github/workflows/create-versioning-pr.yaml
+++ b/.github/workflows/create-versioning-pr.yaml
@@ -9,6 +9,29 @@ on:
     - .github/workflows/create-versioning-pr.yaml
     
 jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    # When a changesets generated Versioning PR is merged, we want to publish the version as a tag
+    if: ${{ startsWith(github.event.commits[0].message, 'Version Kubernetes Agent Chart') }}
+    steps:
+    - uses: actions/checkout@v4
+      with:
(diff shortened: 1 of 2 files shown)


APErebus (approved):
Do we know if this tags the version commit, or the previous one?

scme0:
@APErebus As far as I can tell, the way I've configured it, it should work something like:
1. Merge your PR with a changeset in it
2. GHA creates a Versioning PR
3. Versioning PR is merged manually
4. GHA publishes a tag for that just-merged Versioning PR.

Running the last step locally creates a tag for the commit you currently have checked out, which in this case would be the merged Versioning commit. So theoretically it should tag correctly. We'll have to verify when we add a change with a changeset.

APErebus:
> @APErebus As far as I can tell, the way I've configured it, it should work something like:
> 
> 1. Merge your PR with a changeset in it
> 2. GHA creates a Versioning PR
> 3. Versioning PR is merged manually
> 4. GHA publishes a tag for that just-merged Versioning PR.
> 
> Running the last step locally creates a tag for the commit you currently have checked out, which in this case would be the merged Versioning commit. So theoretically it should tag correctly. We'll have to verify when we add a change with a changeset.

Makes sense! :shipit:

PR #212, opened 2024-06-30
Add Create Release step to GHAs for Kubernetes Agent

scme0:
We've setup the Deploy instance of Octopus to have two channels for the Kubernetes agent helm chart. Unfortunately ARC (Automatic Release Creation) for package pushes doesn't support multiple channels. This means that we are not currently automatically creating releases for pre-release packages.

To fix this, I'm going to turn off ARC and revert to creating releases manually after pushing the package.

👇 No Changeset is required because this is a fix to the Github Actions and not a functional change.


scme0:
I've confirmed that a release was created in deploy instance for this PR which was automatically deployed to Staging 👍 

Screenshot 2024-07-01 at 08 30 04

APErebus (approved):

rain-on (approved):
LGTM

PR #216, opened 2024-07-02
Change integration tests to use Artifactory

APErebus:
We are having issues executing these tests due to the change to remove the Docker Hub rate limiting exemption for azure-hosted resources


MissedTheMark (approved):
LTGM.

PR #220, opened 2024-07-03
Make v1 Helm chart compatible with new Tentacle env variable names

kevjt:
## Background
There have been some [changes](https://github.com/OctopusDeploy/OctopusTentacle/pull/969) to Tentacle to support the installation of the kubernetes agent as a worker. These changes break the contract between the Helm chart and the Tentacle start up script, as the names of some environment variables have changed etc.

We would like to backport some of the v2 Helm chart changes to v1 so that v1 will be able to easily upgrade to newer Tentacle versions going forward. 

## Results
Updating the name of the environment variables set by the Helm chart such that it is compatible with the latest changes to Tentacle.

### DIFF
--- a/.changeset/hot-buckets-dress.md
+++ b/.changeset/hot-buckets-dress.md
@@ -1,0 +1,5 @@
+---
+"kubernetes-agent": patch
+---
+
+Make v1 agent compatible with the latest changes to Tentacle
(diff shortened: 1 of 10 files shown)


APErebus (approved):
Excellent 👍

PR #221, opened 2024-07-04
Update to pnpm/action-setup@v4

APErebus:
We are getting failures with the previous v2.

Comments in an issue said that upgrading to v4 fixes the issue: https://github.com/pnpm/action-setup/issues/135


MissedTheMark (approved):
LGTM

PR #225, opened 2024-07-09
Support configuring script pod resource requirements

APErebus:
A customer appears to be having resource contention with a large number of script pods. They would like to be able to modify the resource requirements of script pods. 

This PR adds a new values node in the `scriptPods` section. These `resources` are then serialized to JSON and passed to the `agent-tentacle` container as an environment variable. This is consumed by the change in https://github.com/OctopusDeploy/OctopusTentacle/pull/977

Tested locally with 
```
--set scriptPods.resources.limits.cpu="150m" `
--set scriptPods.resources.limits.memory="250Mi" `
--set scriptPods.resources.requests.cpu="50m" `
--set scriptPods.resources.requests.memory="150Mi" 
```
resulted in tentacle env var

image

resulted in script pods config

image

### DIFF
--- a/charts/kubernetes-agent/values.yaml
+++ b/charts/kubernetes-agent/values.yaml
@@ -123,6 +123,13 @@ scriptPods:
   # @section -- Script pod values
   disruptionBudgetEnabled: true
 
+  # -- The resource limits and requests assigned to script pod containers
+  # @section -- Script pod values
+  resources: 
+    requests:
+      memory: "100Mi"
+      cpu: "20m"
+
   serviceAccount:
     # -- The name of the service account used for executing script pods
     # @section -- Script pod values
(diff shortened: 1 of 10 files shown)


charts/kubernetes-agent/values.yaml
```diff
@@ -123,6 +123,13 @@ scriptPods:
   # @section -- Script pod values
   disruptionBudgetEnabled: true
 
+  # -- The resource limits and requests assigned to script pod containers
+  # @section -- Script pod values
+  resources: 
+    requests:
+      memory: "100Mi"
+      cpu: "20m"
```
kevjt, line 131: Should the cpu limit be 25m to match the current default given by Tentacle?

APErebus: 🤦 Yep, will fix

liam-mackie (approved):
LGTM! Thanks for the 🍐 on the review.
For posterity - the test should pass once the tentacle PR is merged and the build is available on dockerhub.

PR #229, opened 2024-07-12
Add serviceName to NFS StatefulSet

APErebus:
When using a `StatefulSet` with a headless service, you are supposed to set the `spec.serviceName` value. We weren't.

However, once a stateful set is created, this value can't be modified. As a result. I'm using the `.Release.IsInstall` flag to only set the `spec.serviceName` value when the agent is being created for the first time.

Not setting this value isn't a big issue because we don't have multiple replica's of the NFS pods

### DIFF
--- a/.changeset/dry-feet-add.md
+++ b/.changeset/dry-feet-add.md
@@ -1,0 +1,5 @@
+---
+"kubernetes-agent": patch
+---
+
+Correctly set serviceName on NFS StatefulSet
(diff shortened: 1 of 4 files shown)


liam-mackie (approved):
LGTM! Thanks for adding an install test as well for the current version. 
The only thing I'm not certain of is whether this now breaks installation via ArgoCD, since I'm not certain if argo works with `lookup` functions (see https://github.com/argoproj/argo-cd/issues/5202)

APErebus:
Converting back to draft as we may not want to actually merge this change

APErebus:
Closing as this is not something that has come up as an issue with customers

PR #231, opened 2024-07-15
Tmm/worker perms

rain-on:
Creating a role which is able to modify the local namespace only - and bind it to the serviceAgent when running as a worker.


charts/kubernetes-agent/templates/_helpers.tpl
```diff
@@ -72,6 +72,15 @@ Create the name of the pod cluster role for deleting pods
 {{- printf "%s-delete-role" (include "kubernetes-agent.scriptPodServiceAccountFullName" .) }}
 {{- end }}
 
+
+{{/*
+Create the name of the pod (namespaced) role for creating pods, and updating tentacle
+*/}}
+{{- define "kubernetes-agent.scriptPodWorkerRoleName" -}}
+{{- printf "%s-worker-role" (include "kubernetes-agent.scriptPodServiceAccountFullName" .) }}
```
liam-mackie, line 80: Minor: this doesn't protect against names >63 characters - the trunc is in the `kubernetes-agent.scriptPodServiceAccountFullName` and so we may have names that are invalid here. This is a problem already so happy to ignore for this PR.

charts/kubernetes-agent/templates/pod-rolebindings.yaml
```diff
@@ -19,6 +21,22 @@ roleRef:
   name: {{ $podClusterRoleName }}
```
liam-mackie, line 21: It might be better to switch here and modify the roleref - it makes it much easier to understand what's happening.

charts/kubernetes-agent/templates/pod-roles.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "kubernetes-agent.scriptPodWorkerRoleName" . }}
```
liam-mackie, line 4: Is this always deployed? We may want to consider allowing users to override this.

liam-mackie (approved):
Minor comments, but LGTM :)

charts/kubernetes-agent/templates/pod-roles.yaml
rain-on: Good point - I've wrapped this in a "if worker is enabled" - is that the cleanest way to protect against deploying an unwanted role?

charts/kubernetes-agent/templates/pod-rolebindings.yaml
rain-on: Tried to make that fit - but couldn't so broke out the role-binding to another file

.changeset/nice-tomatoes-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+"kubernetes-agent": patch
```
APErebus, line 2:
```suggestion
"kubernetes-agent": minor
```

APErebus (approved):

charts/kubernetes-agent/templates/pod-roles.yaml
liam-mackie: For this, yeah. Is this role unable to be defined by the user at all?

charts/kubernetes-agent/templates/pod-rolebindings.yaml
liam-mackie: We spoke about this, and decided that this was a much easier and more elegant way to do this. If anyone is curious, the following code should work if you did want to reuse the code:
```
{{- $targetNamespaces := list -}}
{{- if .Values.agent.deploymentTarget.enabled }}
{{- $targetNamespaces := list .Release.Namespace }}
{{- else }}
{{- $targetNamespaces := $.Values.scriptPods.serviceAccount.targetNamespaces }}
{{- end }}
{{- range $targetNamespace := $targetNamespaces }}
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: {{ printf "%s-binding" $podServiceAccountFullName }}
  namespace: {{ $targetNamespace | quote }}
subjects:
- kind: ServiceAccount
  name: {{ $podServiceAccountName }}
  namespace: {{ $.Release.Namespace | quote }}
roleRef:
{{- if .Values.agent.deploymentTarget.enabled }}
  kind: Role
  name: {{ $podWorkerRoleName }}
  apiGroup: rbac.authorization.k8s.io
{{- else }}
  kind: ClusterRole
  name: {{ $podClusterRoleName }}
  apiGroup: rbac.authorization.k8s.io
{{- end}}
{{- end -}}
```

charts/kubernetes-agent/templates/pod-roles.yaml
rain-on: I think no - for now, its implicit.

PR #234, opened 2024-07-23
Fix the mapping of imagePullSecrets to a CSV list

APErebus:
I made a mistake with the `imagePullSecrets` and serializing this value to the CSV list that the tentacle code expects. This was also compounded by the fact the unit tests weren't correctly configuring the `imagePullSecrets` test values.

### DIFF
--- a/charts/kubernetes-agent/templates/_helpers.tpl
+++ b/charts/kubernetes-agent/templates/_helpers.tpl
@@ -103,3 +103,16 @@ The name of the PersistentVolumeClaim to configure
 {{- include "nfs.pvcName" . }}
 {{- end }}
 {{- end }}
+
+{{/* 
+Turns the imagePullSecrets map into a CSV.
+*/}}
+{{- define "kubernetes-agent.imagePullSecretsCsv" -}}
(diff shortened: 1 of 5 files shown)


charts/kubernetes-agent/templates/_helpers.tpl
```diff
@@ -102,4 +102,17 @@ The name of the PersistentVolumeClaim to configure
...
+
+{{/* 
+Turns the imagePullSecrets map into a CSV.
+*/}}
+{{- define "kubernetes-agent.imagePullSecretsCsv" -}}
+{{- if .Values.imagePullSecrets }}
+{{- $imagePullSecretCsv := (first .Values.imagePullSecrets).name }}
+{{- range $i, $val := (rest .Values.imagePullSecrets) }}
+    {{- $imagePullSecretCsv = (printf "%s,%s" $imagePullSecretCsv $val.name) }}
+{{- end }}
```
APErebus, line 115: I feel like there should be a better way to do this, but I couldn't find it

liam-mackie (approved):
LGTM! Thanks for adding a test for the single pull secret scenario :)

PR #236, opened 2024-07-23
Update values file to provide a default image and tag for use during script-pod creation

rain-on:
Currently - a script pod will be either:
* kubernetes-agent-base-tools (by default), or
* an image as defined on the script execution command

When the kuberentes-agent is used as a worker, it is almost guaranteed that the default container will have insufficient tooling for most workflows - thus requiring the customer to define an execution-container on any/all steps which may be sent to a k8s worker.

This change introduces helm values which can be used to define the container used when spawning a script pod.

These values will be passed into the Tentacle containers as environment-variables - and will be dependent on whether the agent is operating as a worker or a deployment target.


charts/kubernetes-agent/values.yaml
```diff
@@ -163,6 +163,18 @@ scriptPods:
     clusterRole:
       rules: []
 
+  deploymentTarget:
```
APErebus, line 166: These are missing comments. See the other image options for examples

rain-on: added comments

APErebus: These are at the wrong level, they should be above the `image` node (see the output)

.changeset/polite-donuts-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+"kubernetes-agent": minor
+---
+
+Allow the script-pod container to be defined in values
```
APErebus, line 5:
```suggestion
Allow the container used by script pods to be modified for both worker and deployment target
```

APErebus (changes requested):

charts/kubernetes-agent/values.yaml
rain-on: resolved

charts/kubernetes-agent/values.yaml
```diff
@@ -163,6 +163,22 @@ scriptPods:
...
+    # -- The repository, pullPolicy & tag to use for the script-pod image used when the cluster is a deploymentTarget
```
APErebus, line 167:
```suggestion
    # -- The repository, pullPolicy & tag to use for the script pod image when the agent is a deployment target
```

charts/kubernetes-agent/values.yaml
```diff
@@ -163,6 +163,22 @@ scriptPods:
...
+    # @section -- Script pod values
+    image:
+      repository: ""
+      pullPolicy: ""
+      tag: ""
+      
+  worker:
+    # -- The repository, pullPolicy & tag to use for the script-pod image used when running as a worker
```
APErebus, line 175:
```suggestion
    # -- The repository, pullPolicy & tag to use for the script pod image when the agent is a worker
```

APErebus (commented):
Couple of minor spelling issues and then we'll need to update the tentacle tag when the tentacle PR is merged

APErebus (approved):
Excellent work mate!

PR #239, opened 2024-07-25
Update Tentacle to fix bug with whitespaces not allowed in target roles

kevjt:
Fixes a bug with which whitespaces in input fields (e.g. target role) were causing the registration to fail due to incorrect tokenisation.

[Tentacle PR](https://github.com/OctopusDeploy/OctopusTentacle/pull/983)


APErebus (approved):
Awesome. Thank you mate 🙇

PR #248, opened 2024-07-29
Add ability to modify agent, nfs and script pod affinity and tolerations

APErebus:
Customers have asked us to allow them to set the pod affinity & tolerations on both the agent pod and script pods.

This PR adds these values to the `agent` section and `scriptPods` section

Shortcut story: [sc-86605]

### DIFF
--- a/charts/kubernetes-agent/templates/tentacle-deployment.yaml
+++ b/charts/kubernetes-agent/templates/tentacle-deployment.yaml
@@ -102,10 +102,16 @@ spec:
             {{- end }}
             - name: "OCTOPUS__K8STENTACLE__PODRESOURCEJSON"
               value: {{ .Values.scriptPods.resources | toJson | quote }}
+            - name: "OCTOPUS__K8STENTACLE__PODAFFINITYJSON"
+              value: {{ .Values.scriptPods.affinity | toJson | quote }}
+            {{- with .Values.scriptPods.tolerations }}
+            - name: "OCTOPUS__K8STENTACLE__PODTOLERATIONSJSON"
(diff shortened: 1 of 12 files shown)


APErebus:
PR still needs to be updated with Tentacle version once built and approved

charts/kubernetes-agent/templates/tentacle-deployment.yaml
```diff
@@ -102,10 +102,16 @@ spec:
             {{- end }}
             - name: "OCTOPUS__K8STENTACLE__PODRESOURCEJSON"
               value: {{ .Values.scriptPods.resources | toJson | quote }}
+            - name: "OCTOPUS__K8STENTACLE__PODAFFINITYJSON"
+              value: {{ .Values.scriptPods.affinity | toJson | quote }}
```
liam-mackie, line 106: If someone overrides the affinity to something empty, should we replace with the default or assume they made the change on purpose?

liam-mackie (approved):
As discussed, we will want to expose the NFS pod's tolerations and affinities as well. Otherwise, LGTM!

charts/kubernetes-agent/templates/tentacle-deployment.yaml
APErebus: 🤔 Good question.

For backwards compatibility, I think we _need_ to have it set the default. Existing users won't have a value in the values.yaml, so we don't want to set the affinity empty in those cases.

Until we hear otherwise, I'm comfortable having it just fallback on the hardcoded defaults

PR #252, opened 2024-07-31
Backport support for changing script pod image, tag and pullPolicy

APErebus:
V2 added support for changing the scriptPods image in #236.

This backports that change to v1. The current tentacle version already supports these tags

### DIFF
--- a/charts/kubernetes-agent/values.yaml
+++ b/charts/kubernetes-agent/values.yaml
@@ -191,6 +191,13 @@ scriptPods:
                 - arm64
                 - amd64
 
+  # -- The repository, pullPolicy & tag to use for the script pod image. If left blank, will use the `octopusdeploy/kubernetes-agent-tools-base` image.
+  # @section -- Script pod values
+  image:
+    repository: ""
+    pullPolicy: ""
+    tag: ""
+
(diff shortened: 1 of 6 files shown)


charts/kubernetes-agent/values.yaml
```diff
@@ -191,6 +191,13 @@ scriptPods:
                 - arm64
                 - amd64
 
+  # -- The repository, pullPolicy & tag to use for the script pod image. If left blank, will use the `octopusdeploy/kubernetes-agent-tools-base` image.
+  # @section -- Script pod values
+  image:
```
liam-mackie, line 196: Will this be a pain since the `image` is in the `deploymentTarget` section in the v2 of the chart?

liam-mackie (approved):
Minor note, but otherwise LGTM!

charts/kubernetes-agent/values.yaml
APErebus: yes, however I didn't want to backport some of the values structure changes. The migration code in V2 should handle this.

APErebus: The forward merge will be a bit fiddly, but should be manageable

PR #256, opened 2024-07-31
Fix GHA to automatically tag commit when versioning PR merged

APErebus:
2 issues:

1. The changeset versioning tag is `xxx@version` where as we have been using `xxx/version`. This cannot be altered in changesets
2. The `git push --follow-tags` didn't work. I've been manually tagging the commits afterwards.

The change is now to just pull the version from the `package.json` and create and push that new tag


kevjt (approved):
We should also probably remove the `publish-version-tags` script from package.json now that it isn't needed.

PR #258, opened 2024-08-01
Add support for setting pods securityContext

APErebus:
Customers have asked if we can run the tentacle application in the container as a non-root user.

This PR adds the ability to set the `securityContext` field on the tentacle pod and script pods via `agent.securityContext` and `scriptPods.securityContext` fields

Shortcut story: [sc-87998]

### DIFF
--- a/.changeset/sharp-pants-attend.md
+++ b/.changeset/sharp-pants-attend.md
@@ -1,0 +1,5 @@
+---
+"kubernetes-agent": minor
+---
+
+Add support for setting pods securityContext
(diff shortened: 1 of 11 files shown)


liam-mackie (approved):
LGTM, though we may want to consider adding a test for `OCTOPUS__K8STENTACLE__PODSECURITYCONTEXTJSON` being set correctly as well.

APErebus:
> LGTM, though we may want to consider adding a test for `OCTOPUS__K8STENTACLE__PODSECURITYCONTEXTJSON` being set correctly as well.

Added in  in 12bcfb0

PR #262, opened 2024-08-02

APErebus:
Main change is:

- OctopusDeploy/helm-charts/pull/258

plus some other improvements to workflows

### DIFF
--- a/.github/workflows/create-versioning-pr.yaml
+++ b/.github/workflows/create-versioning-pr.yaml
@@ -30,8 +30,12 @@ jobs:
         version: 8.15.2
         run_install: true
 
-    - name: Publish Version Tag
-      run: pnpm publish-version-tags
+    - name: 'Tag version'
+      run: |
+        version=$(jq -r .version charts/kubernetes-agent/package.json)
+        tag="kubernetes-agent/$version"
+        git tag $tag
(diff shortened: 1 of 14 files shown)


.github/workflows/create-versioning-pr.yaml
```diff
@@ -30,8 +30,12 @@ jobs:
         version: 8.15.2
         run_install: true
 
-    - name: Publish Version Tag
-      run: pnpm publish-version-tags
+    - name: 'Tag version'
+      run: |
+        version=$(jq -r .version charts/kubernetes-agent/package.json)
+        tag="kubernetes-agent/$version"
+        git tag $tag
```
kevjt, line 37: Will this start pushing tags for the prerelease versions (because the process was previously not working automatically)? Not sure if there are any reasons we shouldn't be pushing alpha tags 🤷

kevjt (approved):
LGTM, just one question

.github/workflows/create-versioning-pr.yaml
APErebus: it will, but I don't think that's a bad idea. Tagging is lightweight enough and it's really just for us to the exact version of the code for a release

PR #264, opened 2024-08-06
Octopus v1

liam-mackie:
This PR is for the v1 of the octopus server helm chart. 
This v1 will be versioned separately to before, and published in GHCR instead of dockerhub.
This has many changes including:
* A faster, easier setup
* Labels that are the same as kubernetes/helm standards
* Ingress for polling tentacles
* Updated documentation (more to come!)


MJRichardson (approved):
Let's do it

.github/workflows/create-versioning-pr.yaml
```diff
@@ -1,20 +1,23 @@
...
+    - charts/octopus-deploy/**
     - .github/workflows/create-versioning-pr.yaml
     
 jobs:
   publish:
     runs-on: ubuntu-latest
 
     # When a changesets generated Versioning PR is merged, we want to publish the version as a tag
-    if: ${{ startsWith(github.event.commits[0].message, 'Version Kubernetes Agent Chart') }}
+    if: ${{ startsWith(github.event.commits[0].message, 'Version Charts') }}
```
APErebus, line 20: The tagging here is only set up for the `kubernetes-agent/{version}`... I _assume_ you want to tag it `octopus-deploy/{version}`?

APErebus (changes requested):
There is going to be some issues with the K8s Agent publishing workflow due to the PR name change

https://github.com/OctopusDeploy/helm-charts/blob/main/.github/workflows/kubernetes-agent-publish-chart.yaml#L118

.github/workflows/create-versioning-pr.yaml
liam-mackie: Yeah - I'll add a step to do so...

liam-mackie: We moved this over to a new workflow and only trigger it when changes are made to the agent. This may fail on the next versioning run when this gets merged, as it'll run since we changed the workflow itself.

APErebus (approved):
Workflow changes look good to me!

PR #265, opened 2024-08-06
Exit pre-release versioning mode to release v2 of the Kubernetes Agent helm chart

kevjt:
This [exits](https://github.com/changesets/changesets/blob/main/docs/prereleases.md) the pre-release mode that is being used to develop v2-alpha.

The next automatic versioning PR will update the version to 2.0.0 and update the changelog.

PR #272, opened 2024-08-08
Run versioning workflow when .changeset files change

APErebus:
So when we exit or enter pre-release, a new versioning PR is created (like #265 )


kevjt (approved):
yes, good point

PR #274, opened 2024-08-08
Script to upgrade from V1 to V2

rain-on:
Temporary PR containing a script to show how we may be able to do the migrations from v1 to v2.


charts/kubernetes-agent/migrations/v1_to_v2/upgrade_v1_to_v2.sh
```diff
@@ -0,0 +1,65 @@
...
+# in the Values file.
+# This script:
+# 1. Extracts values from currently installed kubernetes agent (in json)
+# 2. Transforms the values into the shape expected by a V2 install, and stores in a helm override file using a jq filter
+# 3. Upgrades the helm chart - by resetting the release's values to those in the new chart then applying the previous
+#    current release modified values "over the top" then migrating the moved values
+
+NAMESPACE=octopus-agent-theagent
+RELEASE=theagent
+CHART="oci://docker.packages.octopushq.com/kubernetes-agent"
```
rain-on, line 14: This needs to change to point to the dockerhub version.

PR #286, opened 2024-08-15
Fix permissions when running as worker

rain-on:
It was found that when running as a worker, the clusterrole (used in deployment-target mode) was still attached to the service account being used to run scritps.

This meant that worker-scripts were able affect the greater cluster.

This fix removes the cluster-rolebinding from the serviceAccount when in worker-mode.

image

### DIFF
--- a/charts/kubernetes-agent/tests/pod-clusterbinding_test.yaml
+++ b/charts/kubernetes-agent/tests/pod-clusterbinding_test.yaml
@@ -4,6 +4,7 @@ templates:
 tests:
 - it: "is not created when target namespaces is not empty"
   set:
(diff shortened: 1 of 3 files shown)


charts/kubernetes-agent/tests/pod-clusterbinding_test.yaml
```diff
@@ -4,6 +4,7 @@ templates:
 tests:
 - it: "is not created when target namespaces is not empty"
   set:
+    agent.worker.enabled: false
```
APErebus, line 7: this isn't _strictly_ necessary for the test to pass right?

APErebus (approved):

PR #293, opened 2024-08-22
Create Clusterrole allowing pv actions

rain-on:
This clusterrole is then applied to the service-account via a cluster-role binding such that worker script-pods may operate on persistent-volumes, as is required during a helm upgrade.


APErebus:
@rain-on I think we can close this?

rain-on:
overtaken by events.

PR #294, opened 2024-08-26
Add service account for upgrading

kevjt:
We have identified an issue where the permissions for the scripts pods do not allow for Octopus Server to perform automatic upgrading.

To resolve this, we are adding a new service account with cluster-wide permissions that is only used by Octopus Server to perform automatic upgrades.

This is part of the fix for: https://github.com/OctopusDeploy/Issues/issues/8985


APErebus (approved):
LGTM!

🚀

PR #297, opened 2024-08-26

kevjt:
Includes version `1.16` which adds service account for allowing automatic upgrades via Octopus Server

Includes #294


APErebus (approved):
🚀

Issue #305, opened 2024-08-28
Helm install of ocotpus server fails if trying to set masterKey through values


johatl:
When running the helm deployment with masterKey set, the deployment fails:

`helm upgrade --install --atomic --set octopus.masterKey='*************' --set octopus.databaseConnectionString='dummy' octopus oci://ghcr.io/octopusdeploy/octopusdeploy-helm 
Release "octopus" does not exist. Installing it now.
Pulled: ghcr.io/octopusdeploy/octopusdeploy-helm:1.0.0
Digest: sha256:5510a6b80c0342fb088d4f04140018bdec550bdf177106b73863c126e83dbe77
Error: template: octopusdeploy-helm/templates/NOTES.txt:10:42: executing "octopusdeploy-helm/templates/NOTES.txt" at <or>: wrong number of args for or: want at least 1 got 0`

APErebus:
@liam-mackie Can this issue be closed now?

PR #308, opened 2024-09-10
Fix octopus server notes rendering

liam-mackie:
It was reported in #305 that the notes were failing to render due to an incorrectly written `or` conditional in the notes. This PR fixes this so that master keys are able to be provided now.


APErebus (approved):
👍

PR #310, opened 2024-09-11
Upgrade Kubernetes Tentacle to .NET 8 & Debian 12

kevjt:
## Background

Tentacle has moved from .NET 6.0 to .NET 8.0, and in the process the agent's base image also received an upgrade from Debian 11 to Debian 12. 

Due to a compatibility issue on Debian 12 with older SHA1 certificates (more explanation [here](https://github.com/OctopusDeploy/OctopusTentacle/pull/997)), we will for now release both Debian 11 and Debian 12-based images for each version of the agent.

The plan is to set Debian 12 as the default base image and provide a configuration option in the agent Helm chart, allowing users to switch to Debian 11 if needed.

## Results

- Bump Kubernetes Tentacle image
- A new optional `agent.image.tagSuffix` field has been added to the agent's values configuration, which will be appended to the end of the image tag, separated by a hyphen.

### DIFF
--- a/charts/kubernetes-agent/templates/_helpers.tpl
+++ b/charts/kubernetes-agent/templates/_helpers.tpl
@@ -162,3 +162,22 @@ The Env-var block required to set image name, tag and pullpolicy
   value: {{ .pullPolicy | quote}}
 {{- end }}
 {{- end }}
+
+{{/*
+The base image tag for the agent, without any suffixes.
+Defaults to the Chart Appversion.
+*/}}
+{{- define "kubernetes-agent.imageTag" -}}
+{{- print .Values.agent.image.tag | default .Chart.AppVersion}}
+{{- end }}
+
+{{/*
+The complete image tag for the agent, including any optional suffixes.
+*/}}
+{{- define "kubernetes-agent.fullImageTag" -}}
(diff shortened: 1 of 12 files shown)


APErebus:
We should target this PR against `release/kubernetes-agent/v1` and merge forward

kevjt:
> We should target this PR against `release/kubernetes-agent/v1` and merge forward

Yes you're right. I did branch off v1, but I opened the PR using Git Fork's 'push and create PR' option which only ever targets main by default.

charts/kubernetes-agent/templates/_helpers.tpl
```diff
@@ -162,3 +162,22 @@ The Env-var block required to set image name, tag and pullpolicy
   value: {{ .pullPolicy | quote}}
 {{- end }}
 {{- end }}
+
+{{/*
+The base image tag for the agent, without any suffixes.
+Defaults to the Chart Appversion.
+*/}}
+{{- define "kubernetes-agent.imageTag" -}}
+{{- print .Values.agent.image.tag | default .Chart.AppVersion}}
```
APErebus, line 171: Let's change this to be the full image including the repository

PR #316, opened 2024-09-15
feat: add topology spread constraints to OD StatefulSet

liam-mackie:
This PR adds topology spread constraints, as well as a test for them. This should allow people to spread their server installations over multiple domains much easier. 
An example of how you could use this is as follows:
```
octopus:
  topologySpreadConstraints:
  - maxSkew: 1
    topologyKey: zone
    whenUnsatisfiable: DoNotSchedule
    labelSelector:
      matchLabels:
        app.kubernetes.io/component: octopus-server
```
This would ensure that pods need to be spread evenly (with a maximum skew of 1) over the `zone` topology key.

### DIFF
--- a/.changeset/grumpy-dots-yell.md
+++ b/.changeset/grumpy-dots-yell.md
@@ -1,0 +1,5 @@
+---
+"octopus-deploy": minor
+---
+
+Add topology spread constraints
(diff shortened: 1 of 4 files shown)


APErebus (approved):
Interesting!

PR #318, opened 2024-09-23
Add values migration, tests, and workflow

MissedTheMark:
To allow us to change the structure of our K8s agent/worker's Helm chart values file, we need to be able to migrate user-specified values.  Our [helm-migrate-values ](https://github.com/OctopusDeployLabs/helm-migrate-values) Helm plugin provides that ability, and this PR adds the migration template for migrating from v1.x.x to v2.x.x

This PR also adds a simple test framework for values migrations generally, as well as test data for the v1-v2 migrations to help provide a safety net ensuring that the migrations actually do what we expect.

## To the reviewer
Please take time to ensure that the migration covers what _you_ expect it to. I have done a diff of the v1 and v2 values.yaml files to determine the values that need migration, and it would be good for you to do the same.

Once this is merged, I will create a feature branch instance of the version of Octopus Deploy which supports automatically upgrading using this migration - if you also could test a migration from a v1 chart to v2 on your own cluster via Octopus, it would give us another solid indicator of its correctness. 

I have tested a slightly older version of the migration in Octopus Deploy, and will be testing this one myself as well.


charts/kubernetes-agent/value-migrations-tests/to-v2-requiredValuesOnly-expected.yaml
```diff
@@ -0,0 +1,10 @@
+agent:
+  name: "agent-name"
+  deploymentTarget:
+    enabled: true
+    initial:
+      environments: ["env1", "env2"]
+      tags: ["tag1", "tag2" ]
+  targetEnvironments: null
+  targetName: null
+  targetRoles: null
```
MissedTheMark, line 10: These ensure that the values in the old format are removed.

charts/kubernetes-agent/value-migrations-tests/v1-allValues-initial.yaml
MissedTheMark, line 1: This file provides all the values which can be migrated.

charts/kubernetes-agent/value-migrations-tests/v1-requiredValuesOnly-initial.yaml
MissedTheMark, line 1: This file provides only the mandatory values to migrate.

charts/kubernetes-agent/value-migrations/to-v2.yaml
```diff
@@ -0,0 +1,52 @@
...
+  {{- end}}
+  {{- if .agent.targetTenants }}
+  targetTenants: null
+  {{- end}}
+  {{- if .agent.targetTenantedDeploymentParticipation }}
+  targetTenantedDeploymentParticipation: null
+  {{- end}}
+  {{- if .agent.scriptPods.image }}
+  image: null
+  {{- end}}
```
APErebus, line 37: What's this for?
Should this be `.agent.image`?

MissedTheMark: Good spot, that's a section that I straight up should have removed.  It's actually covered by the `scriptPods:image:null` bit below.

charts/kubernetes-agent/value-migrations-tests/go.mod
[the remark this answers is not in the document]
MissedTheMark: Just added them to the `.helmignore`.

kevjt (approved):
LGTM

PR #322, opened 2024-10-02
Update Tentacle version to include the workaround for ensuring the agent uses the latest available tools image

kevjt:
Details about the what and why are in the description of https://github.com/OctopusDeploy/OctopusTentacle/pull/1010


APErebus (approved):
nice work 🚀

PR #335, opened 2024-11-25
Provide credentials for using during helm upgrade.

rain-on:
Helm has a config file which contains credentials for the registries to which it is logged in (it can otherwise fallback on docker config).

This file is the same format as a docker config file.

```
{
     "auths": {
                     "<registry>": {
                              "auth":"dXNlcm5hbWU6cGFzc3dvcmQ="
                      }
       }
}
```
Where registry (for dockerhub) should be "https://index.docker.io/v1".
And "auth" is a basic auth string - i.e. a base64 encoding of `<username>:<password>`

While there are many more fields in this file, this is sufficient to get our e2e tests passing reliably.


charts/kubernetes-agent/values.yaml
```diff
@@ -331,6 +331,12 @@ persistence:
         pullPolicy: IfNotPresent
         tag: "0.2.0"
 
+upgrade:
+  username: ""
+  password: ""
+  registry: ""
+    
+
```
liam-mackie:
```suggestion
upgrade:
  dockerAuth:
    username: ""
    password: ""
    registry: ""
```
Consider changing the root from `upgrade` - it's not very descriptive.

liam-mackie (approved):
LGTM with minor comments

.changeset/two-masks-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+"kubernetes-agent": minor`
+---
+
+Can supply credentials for use during agent upgrades
```
APErebus, line 5:
```suggestion
Add `agent.upgrade.dockerAuth` to support authenticating the docker calls when performing an automatic upgrade via Octopus Server.
```

APErebus (approved):
Just 1 minor suggestion for the release notes, but LGTM!

PR #338, opened 2024-11-28
Update Chart.yaml AppVersion to match default Tentacle version

rain-on:
The Chart.yaml's AppVersion was not updated to match the default Tentacle vesrion. This in turn causs E2E tests in OctopusServer to fail.


kevjt (approved):
LGTM

PR #343, opened 2024-12-02
Update kubernetes-agent-tentacle to enable pod log encryption

APErebus:
Updates the tentacle container to include the changes https://github.com/OctopusDeploy/OctopusTentacle/pull/1047 and https://github.com/OctopusDeploy/OctopusTentacle/pull/1048 

As a result, all script pod logs will now be encrypted with a per-pod encryption key to avoid leaking sensitive values

Shortcut story: [sc-98480]

PR #345, opened 2024-12-02
Forward merge release/kubernetes-agent/v1

APErebus:
Forward merge `release/kubernetes-agent/v1`.

This includes #343 which is the enables pod log encryption

It also includes the commit for #341 which was a backport of the upgrade credentials change


rain-on (approved):
LGTM

PR #354, opened 2024-12-05
feat: add kubernetes monitor sub-chart

liam-mackie:
This PR adds the kubernetes monitor to the agent as a helm dependency, and defaults the installation to `false`. This allows us to (in future) allow anyone using the agent to also enable the monitor.


.changeset/slimy-rules-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+"kubernetes-agent": minor
```
APErebus, line 2: Kinda devils advocate, but should this be a `major` bump?

charts/kubernetes-agent/values.yaml
```diff
@@ -339,6 +339,12 @@ persistence:
         pullPolicy: IfNotPresent
         tag: "0.2.0"
 
+ # Used to enable the Kubernetes Monitor
+# Do not use this, as it is not supported at the moment
+# @ignored
+kubernetes-monitor-chart:
```
APErebus, line 345: could just be `kubernetesMonitor`?
The fact it's a chart is kinda irrelevant for the values?

APErebus (approved):

.changeset/slimy-rules-bow.md
liam-mackie: It's backwards compatible, but adding new functionality, so left it as a `minor` according to SemVer.

charts/kubernetes-agent/Chart.lock
```diff
@@ -0,0 +1,6 @@
+dependencies:
+- name: kubernetes-monitor-chart
+  repository: oci://docker.io/octopusdeploy
```
APErebus, line 3: Is it possible to change the repo this chart is pulled from? (say if I want to pull from Artifactory?)

charts/kubernetes-agent/values.yaml
liam-mackie: When setting values, the sub-charts name is the root of the values for that particular chart. This allows us to expose (in future) the documentation for the subchart in our readme. If we use `kubernetesMonitor`, we'd need to name the chart `kubernetesMonitor`, which clashes with the _actual_ image 😅

.changeset/slimy-rules-bow.md
APErebus: yep, happy with that. Just wanted to ask the question :)

charts/kubernetes-agent/values.yaml
APErebus: gotcha 👍

charts/kubernetes-agent/Chart.lock
APErebus: kk, we can always change it if we need to in the future 👍

charts/kubernetes-agent/values.yaml
eddymoulton: You can set an `alias` in `Chart.yaml` to change this.
Unsure if we should or not. It cleans this up but makes it a little more confusing.

Considering we are packaging everything ourselves and don't currently have any expectation that others will do the same without reference, I'm kind of leaning towards yes.

eddymoulton (approved):
Couple of comments for your discretion, but I'm happy with this

charts/kubernetes-agent/templates/pre-install-tentacle-role.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ printf "%s-role" (include "kubernetes-agent.preinstall.serviceAccountName" .) }}
+  namespace: {{ .Release.Namespace | quote }}
+  annotations:
+    {{- include "kubernetes-agent.hookAnnotationsWithSuccessDelete" . | nindent 4 }}
+rules:
+- apiGroups: ["*"]
+  resources: ["pods", "pods/log", "configmaps", "secrets", "events"]
```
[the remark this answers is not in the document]
liam-mackie: I think initially, we needed these because we started the watcher. We probably don't anymore, as the `register` command only works with configmaps and secrets. I'll test it out!

charts/kubernetes-agent/values.yaml
liam-mackie: I'm happy to use an alias (TIL), but is it worth it? Happy to discuss.

eddymoulton: Probably not 😅

charts/kubernetes-agent/templates/pre-install-tentacle-role.yaml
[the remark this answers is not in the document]
liam-mackie: Works a treat now - removed!

PR #357, opened 2024-12-10
Update octopus-publish-chart.yml

zentron:
`upload-artifact@v3` is being [deprecated](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/)


APErebus (changes requested):
~@zentron can you remake this against the `release/kubernetes-agent/v1` branch and then forward merge that branch to `main`? 🙇 ~

APErebus (approved):
I didn't realise this was for the octopus chart (not the agent chart), which doesn't exit in V1, approving

PR #363, opened 2024-12-17
Don't include empty worker pools env var if not a worker

APErebus:
Small bugfix

When running as a deployment target, we were including an empty `WorkerPools` env var in the tentacle deployment.


liam-mackie (approved):
LGTM!

PR #374, opened 2025-01-16
updated app version to 2024.4

benPearce1:


APErebus:
You will need a changeset to trigger the versioning and release process

liam-mackie (approved):
The change itself is good, but needs a changeset as @APErebus mentioned.

yarn.lock
APErebus, line 1: typically we use pnpm/npm... Not sure if we want to handle more than 1?

.changeset/breezy-coats-count.md
```diff
@@ -0,0 +1,5 @@
+---
+"octopus-deploy": major
```
APErebus, line 2: I'm ok with this. @liam-mackie ?

.changeset/breezy-coats-count.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Bumped octopus server version
```
APErebus, line 5: Nit: can we make this more descriptive, something like `Updated installed Octopus Server version to `2024.4`.

APErebus (approved):
@liam-mackie If people install this new major version, will this upgrade their server version? I assume we use the app version for the tag, so it'll upgrade them?

.changeset/breezy-coats-count.md
liam-mackie: This should be a minor bump IMO, as there is nothing incompatible with the current version.

liam-mackie:
> @liam-mackie If people install this new major version, will this upgrade their server version? I assume we use the app version for the tag, so it'll upgrade them?

Yes, if people install this new version, it will upgrade their server version. I don't think this is a bad thing, and we have previously updated the version this chart installs.

benPearce1:
replaced by #375

APErebus:
Closing as it's been replaced by #375

PR #380, opened 2025-02-21
Use corepack version with pnpm

liam-mackie:
This removes the constrained version and updates it to use corepack, so that the pnpm version we use is able to be standardised for everyone as well as our build system.


eddymoulton (approved):
LGTM

PR #409, opened 2025-03-26

APErebus:
Includes changes added in #407

### DIFF
--- a/.changeset/lemon-sheep-punch.md
+++ b/.changeset/lemon-sheep-punch.md
@@ -1,0 +1,5 @@
+---
+"kubernetes-agent": minor
+---
+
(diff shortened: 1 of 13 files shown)


.changeset/lemon-sheep-punch.md
NickJosevski, line 1: what's the significance/need for the codename?
how does it relate to `bullseye-slim` ?

APErebus: it's just a randomly generated file name by the changesets cli

APErebus: The file is removed when the versioning PR comes along (example #408)

NickJosevski (approved):

PR #420, opened 2025-04-10
Add ability to bind PVC to a specific PV via volumeName

APErebus:
We have had requests from customers to be able to specify a specific `volumeName` when using a custom PVC

### DIFF
--- a/charts/kubernetes-agent/templates/tentacle-pvc.yaml
+++ b/charts/kubernetes-agent/templates/tentacle-pvc.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.persistence.storageClassName }}
+{{- if (include "kubernetes-agent.useCustomPvc" .) }}
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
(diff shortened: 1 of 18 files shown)


charts/kubernetes-agent/templates/tentacle-pvc.yaml
```diff
@@ -8,7 +8,12 @@ metadata:
 spec:
   accessModes:
     - ReadWriteMany
-  storageClassName: {{ .Values.persistence.storageClassName | quote }}
+  {{- with .Values.persistence.storageClassName }}
+  storageClassName: {{ . | quote }}
```
liam-mackie, line 12: The customer will still need to define a storage class name to allow the PVC/PV to bind correctly if they provide a volumeName. The accessModes would also need to match. We _may_ be able to set the storageClassName to `""` to allow it to bind, but I'm not sure. Let me check the code.

liam-mackie (changes requested):

charts/kubernetes-agent/templates/tentacle-pvc.yaml
APErebus: So do we need to also expose the `accessModes`?

liam-mackie: It's probably worth talking about whether we do or not

liam-mackie (approved):
LGTM!

Issue #419, opened 2025-04-10
Deployment not working when setting `createSecrets` to `false`


fad3t:
Good morning all,

I noticed there are issues when `createSecrets` is set to false.
- Helm complains the database connection string is not set, most likely because the validation (L30) is done before the condition (L40) in [secrets.yaml](https://github.com/OctopusDeploy/helm-charts/blob/main/charts/octopus-deploy/templates/secrets.yaml#L30). Workaround here was to set a dummy connection string to pass the validation, even if it's not used at all (because `createSecrets` is false, and we create the secret ourselves using ESO).
- Templating of `NOTES.txt` fails because `masterKey` and `password` are not set. But in the case where we create the secrets ourselves, there are no "`generatedSecrets`". Again, workaround was to set dummy values there to get past the error.
- In the template of the StatefulSet, there's a [condition](https://github.com/OctopusDeploy/helm-charts/blob/main/charts/octopus-deploy/templates/statefulset.yaml#L96) for the licenseKey that is never met if `createSecrets` is set to false. Had to put a dummy value for that one too.

I'll try to prepare a PR in the coming weeks to get those fixed, unless someone wants to address this before.

Cheers,
Fred

PR #422, opened 2025-04-22
fix: update conditions to work when createSecrets is set to false

fad3t:
Hello,

This PR updates several conditions in the Octopus Deploy chart to support use cases where `createSecrets` is set to false.

This fixes issue #419.

Didn't find any contribution guidelines, let me know if anything is missing.

Fred


tleed5 (approved):
Changes look good to me, I tested it locally and everything works as expected. In terms of anything missing I think you have got it all covered(changeset and version bump).

Thank you for your contribution and also your patience while we reviewed this 😄

PR #434, opened 2025-05-28
Fix hook annotations

liam-mackie:
This PR fixes hook annotations, which were accidentally incorrect, leading to resources that hung around after successful install.


APErebus (approved):
Awesome. Thanks mate

PR #449, opened 2025-07-13
feat: add extra env variables to octopus server chart

liam-mackie:
This PR adds the ability to define extra environment variables in the octopus server helm chart, as well as adding tests to validate its use. Note that duplicating environment variables already existing in the chart may lead to undefined behaviour (likely a failure to apply in Kubernetes).


zentron (approved):
lgtm

PR #474, opened 2025-09-11
fix: allow pvc for server-log-volume to use global storage class name

laurenlavin:
### Server log pvc not using global.storageClass
Previously the server-log PVC omitted storageClassName when no explicit
octopus.serverLogVolume.storageClassName was provided, ignoring the
global override and leaving the claim unprovisioned on clusters without
a default StorageClass.

- StatefulSet: add fallback to $.Values.global.storageClass for
  server-log-volume.storageClassName when unset
- Preserve "-" sentinel (renders storageClassName: "")
- Use root $.Values and quote the value to avoid YAML issues

#### Before
```
helm template test . --set global.storageClass=cinder --set mssql.enabled="true"
...
  volumeClaimTemplates:
    - metadata:
        name: server-log-volume
      spec:
        accessModes: [ "ReadWriteOnce" ]
        resources:
          requests:
            storage: 200Mi
```

#### After
```
  volumeClaimTemplates:
    - metadata:
        name: server-log-volume
      spec:
        accessModes: [ "ReadWriteOnce" ]
        storageClassName: "cinder"
        resources:
          requests:
            storage: 200Mi
 ```

### DIFF
--- a/.changeset/solid-dryers-join.md
+++ b/.changeset/solid-dryers-join.md
@@ -1,0 +1,9 @@
+---
+"octopus-deploy": minor
+---
+
+Fix server-log-volume to use global storage class name
+
+Previously the server-log PVC omitted storageClassName when no explicit
+octopus.serverLogVolume.storageClassName was provided, ignoring the
+global override.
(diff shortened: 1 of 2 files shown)


APErebus:
hi @laurenlavin, thanks for the PR!

Can I get you to add a changeset to your PR so it version bumps the chart?

liam-mackie (approved):
Thanks for the improvement, this looks good!

PR #483, opened 2025-10-02
Fix pod disruption budget

liam-mackie:
[sc-121238]

Previously, we used `minUnavailable` set to `0` to ensure that no pods created by the agent would be removed.  
Unfortunately, since we never implemented a scale resource, this never worked correctly. This PR changes the minAvailable to a high integer value `8888`. This is arbitrary, but should be more than customers will use.  
This should ensure that the PDB fully applies to resources.

[For reference, this is the section of the disruption controller reconciliation that handles this](https://github.com/kubernetes/kubernetes/blob/b587977f7c8fc7ed4bdc5ed86afa9f1dd924667b/pkg/controller/disruption/disruption.go#L808-L848)

### DIFF
--- a/charts/kubernetes-agent/templates/script-pdb.yaml
+++ b/charts/kubernetes-agent/templates/script-pdb.yaml
@@ -5,7 +5,7 @@ metadata:
   name: script-pod-disruption-budget
   namespace: {{ .Release.Namespace }}
 spec:
-  maxUnavailable: 0
+  minAvailable: 8888
   selector:
     matchExpressions:
     - { key: octopus.com/scriptTicketId, operator: Exists }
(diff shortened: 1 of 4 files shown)


APErebus (approved):
BIGGER NUMBERS! 🚀

charts/kubernetes-agent/templates/script-pdb.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: script-pod-disruption-budget
   namespace: {{ .Release.Namespace }}
 spec:
-  maxUnavailable: 0
+  minAvailable: 8888
```
NickJosevski, line 8: Did we consider `8889`?

PR #537, opened 2026-04-15
fix: Update grpc port name to stop istio interception

liam-mackie:
When running Octopus Server with istio, using the `grpc` port name prompts istio to attempt to intercept the traffic for mTLS, causing issues. This change prevents the mesh from intercepting the traffic erroneously.


flin-8 (approved):
LGTM

PR #540, opened 2026-04-16
fix: remove unstable selector labels in mssql and octopus statefulsets

liam-mackie:
## Fixed issue

`matchLabels` (which is immutable) included label `helm.sh/chart` had dynamic value. That was preventing chart from upgrading

```
kind: StatefulSet
metadata:
  name: octopus-deploy
spec:
  selector:
    matchLabels:
      app.kubernetes.io/component: octopus-server
      app.kubernetes.io/instance: octopus-deploy
      app.kubernetes.io/managed-by: Helm
      app.kubernetes.io/name: octopus-deploy
      helm.sh/chart: octopusdeploy-helm-1.12.0 # this is dynamic value that causing issues during helm upgrade command
```

### DIFF
--- a/charts/octopus-deploy/templates/pre-upgrade-hook.yaml
+++ b/charts/octopus-deploy/templates/pre-upgrade-hook.yaml
@@ -22,7 +22,8 @@ metadata:
 rules:
   - apiGroups: ["apps"]
     resources: ["statefulsets"]
-    verbs: ["get", "delete"]
+    namespace: {{ .Release.Namespace }}
+    verbs: ["list", "delete"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
--- a/charts/octopus-deploy/templates/pre-upgrade-hook.yaml
+++ b/charts/octopus-deploy/templates/pre-upgrade-hook.yaml
@@ -61,9 +61,17 @@ spec:
     spec:
       serviceAccountName: {{ template "octopus.fullname" . }}-pre-upgrade
       restartPolicy: Never
+      {{- with .Values.octopus.podSecurityContext }}
+      securityContext:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
         - name: pre-upgrade
           image: bitnami/kubectl:latest
+          {{- with .Values.octopus.containerSecurityContext }}
+          securityContext:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           command:
             - /bin/sh
             - -c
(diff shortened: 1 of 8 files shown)


charts/octopus-deploy/templates/pre-upgrade-hook.yaml
```diff
@@ -0,0 +1,83 @@
...
+    spec:
+      serviceAccountName: {{ template "octopus.fullname" . }}-pre-upgrade
+      restartPolicy: Never
+      {{- with .Values.octopus.podSecurityContext }}
+      securityContext:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+        - name: pre-upgrade
+          image: bitnami/kubectl:latest
```
philipkotliyakov, line 71: Wouldnt it be safer to use a specific tag rather than :latest?

charts/octopus-deploy/templates/pre-upgrade-hook.yaml
```diff
@@ -0,0 +1,83 @@
...
+kind: Role
+metadata:
+  name: {{ template "octopus.fullname" . }}-pre-upgrade
+  labels:
+    {{- include "labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+rules:
```
philipkotliyakov, line 23: I would add a way to disable this hook (and the extra role/rolebinding/service account) since once the user upgrades and all the stateful sets are recreated with the correct labels, there is no longer need to run the hook.

charts/octopus-deploy/templates/pre-upgrade-hook.yaml
oleksandr-codefresh: looks like latest still better.

bitnami uses sha for tag names, which not readable and also the oldest image in registry 2 was created months ago. Maybe they have some retention.

charts/octopus-deploy/templates/pre-upgrade-hook.yaml
oleksandr-codefresh: what default behaviour u suggest?
 
- `enabled: false` - that makes things inconvenient, as requires user to read docs once meet issue and we don't know from which version to which they gonna update.

- `enabled: true` - I guess in most cases users also won't bother about having this extra value

oleksandr-codefresh: added 
```
hooks:
  enableSelectorLabelsFix: true
```

oleksandr-codefresh (approved):

philipkotliyakov (approved):

PR #545, opened 2026-04-20
Update README with Git resources and storage guidance

zentron:
Added guidelines for using Git resources with Octopus in High Availability mode and storage recommendations.


APErebus (approved):

liam-mackie (approved):
LGTM, though we should consider adding this to the chart by default, then just documenting ways to customise it and why it's there

PR #548, opened 2026-04-22
Simplify to-v3 migration

APErebus:
We aren't making changes to the `persistence.storageClassName` or `persistence.volumeName`, so we don't need to include them in the migration

Fixes MD-1791


liam-mackie (approved):
LGTM, same as what we already discussed

PR #550, opened 2026-04-23
Always set accessModes in migration

APErebus:
When running the values file migration, we want to make sure the `ReadWriteMany` access mode is explicitly set to ensure existing behaviour is maintained

Also updates the migration documentation

### DIFF
--- a/charts/kubernetes-agent/migrations.md
+++ b/charts/kubernetes-agent/migrations.md
@@ -5,6 +5,33 @@ the chart may be upgraded:
 * Minor upgrade (\*.1.\*) - backwards compatible enhancement, upgrade can be executed without additional changes (new version adds new capabilities)
 * Major upgrade (1.\*.\*) - breaking-change (eg Values file structure changes), will require user-input to perform upgrade
 
+## Version 2 --> 3
+
+### Overview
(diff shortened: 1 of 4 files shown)


charts/kubernetes-agent/migrations.md
```diff
@@ -5,6 +5,33 @@ the chart may be upgraded:
...
+A side-effect of the change is that by default, the v3 agent no longer scales horizontally across nodes, but schedules the script pods on the same node as the agent/tentacle pod.
+
+The changes to the `values.yaml` file are the following additions:
+
+| Value name | Value | Comment |
+|--|--|--|
+| `persistence.nfs.enabled` | `false` | Indicates if the NFS pod should be used |
+| `persistence.accessModes` | `["ReadWriteOnce"]` | Sets the access mode used for the PVC |
+
+Kubernetes agents upgraded via Octopus Server will automatically continue to use the existing NFS-backed storage with no changes.
```
liam-mackie, line 25: Should we note that we'll be deprecating this in a future version of the agent?

liam-mackie (approved):
LGTM with a minor note that we can look at later

PR #555, opened 2026-04-28
Update tentacle to 9.1.3835 to fix node affinity issue

APErebus:
An issue was unfortunately identified with the Kubernetes agent `2.38.0`, where script pods were being assigned a node affinity arch of `x64`, not `amd64`.

This was due to the change in https://github.com/OctopusDeploy/OctopusTentacle/pull/1141 coupled with a bug in Octopus Server.

As the fix in Octopus Server will take longer to rollout, we are doing a quick fix in tentacle to coerce any `x64` architectures to `amd64`.

Related to MD-1809

### DIFF
--- a/.changeset/slimy-rats-play.md
+++ b/.changeset/slimy-rats-play.md
@@ -1,0 +1,5 @@
+---
+"kubernetes-agent": patch
+---
+
+Fixed an issue where script pods were being assigned a node affinity of x64, not amd64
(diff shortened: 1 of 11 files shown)


NickJosevski (approved):
version bump is consistent, here's hoping that is the number

APErebus:
> version bump is consistent, here's hoping that is the number

narrator: it was

PR #562, opened 2026-04-29
Document deprecation of in-cluster NFS server

APErebus:
Add deprecation notice for in-cluster NFS server.

### DIFF
--- a/charts/kubernetes-agent/templates/NOTES.txt
+++ b/charts/kubernetes-agent/templates/NOTES.txt
@@ -1,11 +1,11 @@
 This is the official Helm chart for the Octopus Deploy Kubernetes Agent.
 
 You can read more on how to use this at https://octopus.com/docs/infrastructure/deployment-targets/kubernetes/kubernetes-agent
+{{ if .Values.persistence.nfs.enabled }}
 
-{{- if (include "nfs.enabled" .) }}
 ---
 Deprecation Notice
 ---
 The built-in NFS server is deprecated and will be removed in a future major release.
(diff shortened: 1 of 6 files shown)


charts/kubernetes-agent/templates/NOTES.txt
```diff
@@ -1,3 +1,11 @@
...
\ No newline at end of file
+You can read more on how to use this at https://octopus.com/docs/infrastructure/deployment-targets/kubernetes/kubernetes-agent
+
+{{- if (include "nfs.enabled" .) }}
+---
+Deprecation Notice
+---
+The built-in NFS server is deprecated and will be removed in a future major release.
+See https://github.com/OctopusDeploy/helm-charts/issues/554
+{{- end }}
```
APErebus: I haven't tested this, just need to see

liam-mackie: You can test this in a helm unit test - might be good to add a test?

liam-mackie (approved):
LGTM. Would prefer a test and a call out in the README if possible, but those are additive.

APErebus:
> LGTM. Would prefer a test and a call out in the README if possible, but those are additive.

Yeah, I haven't got there yet 😅

charts/kubernetes-agent/templates/NOTES.txt
APErebus: yeah, I was going to, just was doing 5 other things

APErebus:
@liam-mackie I've added a test and update the readme with a clearer deprecation notice

PR #563, opened 2026-04-29
Fix README links for version 2 of kubernetes-agent

APErebus:
Updated README links for version 2 to point to the correct README file.


flin-8 (approved):
LGTM

PR #591, opened 2026-05-12
Fix github action to properly filter v* release branches

liam-mackie:
This fixes an issue in the build.


APErebus (approved):
I doth approve. Thanks!

PR #608, opened 2026-05-26
feat: add liveness probe support for polling Tentacles

ATGardner:
Introduce a liveness probe for polling Kubernetes Agents using the `tentacle livez` exec command. This detects when the "pod is Running but the Tentacle process is stuck" due to deadlocks, GC stalls, or frozen polling threads by checking the freshness of a heartbeat file written by the agent. The feature is enabled when
`agent.serverCommsAddress` or `agent.serverCommsAddresses` is set (polling mode) and requires Tentacle image version >= 9.2.3963. All configuration options are exposed under `agent.livenessProbe.*`.

# Description

Implements the chart-side half of [YOS-338](https://linear.app/octopus/issue/YOS-338). Depends on the Tentacle product change (commit f0d67351) shipping in image tag `<tag>`.

# Pre-requisites

- [ ] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [ ] I have added a changeset with an appropriate customer facing description.
- [ ] I have considered appropriate testing for my change.
- [ ] This PR affects all release versions and will need to be forward merged.
  - See the [documentation](https://github.com/OctopusDeploy/helm-charts/blob/main/charts/kubernetes-agent/docs/forward-merging-release-branches.md) if unsure of the process.

PR #612, opened 2026-06-02
Forward merge from release/kubernetes-agent/v2

ATGardner:
contains #608


oleksandr-codefresh (approved):

.changeset/yos-338-polling-tentacle-liveness-probe.md
```diff
@@ -0,0 +1,5 @@
+---
+"kubernetes-agent": minor
+---
+
+Forward merge changes from [2.42.0](#2.42.0)
```
APErebus: Just a small thing. The format of the link anchor needs to not have any `.` in it, so they this would be `#2420`
```suggestion
Forward merge changes from [2.42.0](#2420)
```

APErebus (commented):
1 minor think that needs fixing

PR #637, opened 2026-06-23
Fix Chart.yaml app version and add new test

APErebus:
# Description

In #627 we updated the `.agent.image.tag` value in `values.yaml`, but didn't update the corresponding `.appVersion` value in `Chart.yaml`.

This is needed as we have tests in Octopus Deploy that assert on this.

I have fixed this and added a new check during the GHA test run to verify these are the same value

This also fixes up all the test snapshots and the readme

# Pre-requisites

- [ ] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [ ] I have added a changeset with an appropriate customer facing description.
- [x] This PR affects all release versions and will need to be forward merged.

### DIFF
--- a/charts/kubernetes-agent/Chart.yaml
+++ b/charts/kubernetes-agent/Chart.yaml
@@ -17,4 +17,4 @@ dependencies:
 type: application
 version: "2.45.0"
 # This version number should be the same as the agent.image.tag value as this is the primary application version
-appVersion: "9.2.4000"
+appVersion: "9.2.4125"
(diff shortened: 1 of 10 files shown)


charts/kubernetes-agent/Chart.yaml
```diff
@@ -17,4 +17,4 @@ dependencies:
 type: application
 version: "2.45.0"
 # This version number should be the same as the agent.image.tag value as this is the primary application version
-appVersion: "9.2.4000"
+appVersion: "9.2.4125"
```
APErebus, line 20: This is the value that was not correctly updated in #627

liam-mackie (approved):

PR #724, opened 2026-08-27
update migration docs after customer feedback

liam-mackie:
# Description

This change makes sure we're explicit that existing installations, and installations that use a custom ReadWriteMany storage class, aren't affected by this change.

### DIFF
--- a/charts/kubernetes-agent/migrations.md
+++ b/charts/kubernetes-agent/migrations.md
@@ -1,6 +1,8 @@
 # Migrations
+
 The Kubernetes Agent adheres to the principles of "Semantic Versioning" to define its version, this implies there are three ways
 the chart may be upgraded:
+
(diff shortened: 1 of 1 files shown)


charts/kubernetes-agent/migrations.md
```diff
@@ -11,6 +13,8 @@ the chart may be upgraded:
 
 Version 3 of the Kubernetes agent changes the default storage from the in-cluster backed NFS server utilizing `ReadWriteMany`, to a single node `ReadWriteOnce` utilising the default storage class of the cluster.
 
+This change only affects *new* installations of the Kubernetes agent that don't use a custom `ReadWriteMany` storage class.
```
liam-mackie, line 16: This is the only change - the rest is automatic linting (that I think might be good to include)

APErebus (approved):

PR #732, opened 2026-09-01
Bump Kubernetes Monitor chart to get rehydration fix

### LINKED ISSUE #10207: Child resources deleted while the monitor is offline linger in live status until the next periodic snapshot
[## Summary

When the Kubernetes monitor starts up it is rehydrated with the set of resources Octopus expects it to watch. That rehydration was additive: the monitor was told what to start watching, but never told what to stop watching.

As a result, resources that disappeared from the cluster while the monitor was offline stayed visible in the application's live status after the monitor came back, until the next periodic snapshot cleared them (up to 5 minutes later, per the `monitor-period` default).

Affects the `kubernetes-monitor-chart` sub-chart of the `kubernetes-agent` chart at versions **0.39.0 and earlier** (monitor application 0.29.2 and earlier).

## Steps to reproduce

1. Deploy an application through Octopus that produces child resources (a Deployment, so that ReplicaSets and Pods are tracked as children), and confirm the children appear in live status.
2. Scale the Kubernetes monitor deployment to 0 replicas.
3. Delete some of those children from the cluster — for example scale the Deployment down, or delete its Pods.
4. Scale the monitor back up to 1 replica and wait for it to connect.
5. Look at the application's live status straight away.

## Expected behaviour

Once the monitor has reconnected and reported, live status reflects the current contents of the cluster.

## Actual behaviour

The deleted children are still listed. They are only removed when the monitor's next periodic snapshot lands, so live status can show resources that no longer exist for several minutes after the monitor is demonstrably healthy and connected.

## Cause

Rehydration on connection only ever added to the monitor's watch list. There was no way for Octopus to say "this is the complete list, forget anything else", so records for resources Octopus had stopped tracking were left in place and could only be reconciled by the timer-driven snapshot.

## Fix

The monitor now understands a complete desired-resource list from Octopus and answers it with a full replacement of the monitored resources for that application instance, pruning anything Octopus is no longer tracking, at the moment it connects rather than on the next tick.

This depends on a version of Octopus Server that sends the complete list on connection. Against an older Server the monitor behaves as it did before, and stale records are still cleared by the periodic snapshot.

## Resolution

Fixed in `kubernetes-monitor-chart` **0.40.0** (monitor application 0.30.0).
](https://github.com/OctopusDeploy/octopus-argocd-gateway/actions/runs/33459318924/job/99705936841?pr=274)

flin-8:
# Description

Fixes https://github.com/OctopusDeploy/Issues/issues/10207

# Pre-requisites

- [x] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [x] I have added a changeset with an appropriate customer facing description.
- [x] This PR affects all release versions and will need to be forward merged.

### DIFF
--- a/.changeset/swift-lions-work.md
+++ b/.changeset/swift-lions-work.md
@@ -1,0 +1,5 @@
+---
+"kubernetes-agent": patch
+---
+
+Bump Kubernetes Monitor chart to get rehydration fix
--- a/.changeset/swift-lions-work.md
+++ b/.changeset/swift-lions-work.md
@@ -2,4 +2,4 @@
 "kubernetes-agent": patch
 ---
 
-Bump Kubernetes Monitor chart to 0.40.0
+Bump Kubernetes Monitor chart to 0.40.0. This fixes a bug where child resources deleted while the monitor is offline linger in live status until the next periodic snapshot. See https://github.com/OctopusDeploy/helm-charts/issues/731 for more details.
--- a/.changeset/swift-lions-work.md
+++ b/.changeset/swift-lions-work.md
@@ -2,4 +2,8 @@
 "kubernetes-agent": patch
 ---
 
-Bump Kubernetes Monitor chart to 0.40.0. This fixes a bug where child resources deleted while the monitor is offline linger in live status until the next periodic snapshot. See https://github.com/OctopusDeploy/helm-charts/issues/731 for more details.
+Bump Kubernetes Monitor chart to 0.40.0.
+
+- Fixes a bug where child resources deleted while the monitor is offline linger in live status until the next periodic snapshot
+- Adds a new `ReplaceDesiredResourcesCommand` for Octopus Server to send all desired resources together when monitor connects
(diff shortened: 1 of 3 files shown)


.changeset/swift-lions-work.md
eddymoulton, line 1: I like to add something to explain what changed in the monitor since still don't have that public

eddymoulton (approved):

.changeset/swift-lions-work.md
flin-8: Better?

eddymoulton: Yeah this works, simliar to other change logs we've done.

[This is what I was thinking though](https://github.com/OctopusDeploy/helm-charts/blob/main/charts/kubernetes-agent/CHANGELOG.md#2270). Less paragraph and more dot points

PR #740, opened 2026-09-03
feat: add readiness probe for polling Tentacles

zentron:
# Description

Polling Tentacles expose no port to check directly, so readiness is determined by an exec probe (`ps aux | grep -v grep | grep -q Tentacle`) confirming the Tentacle process is running. All knobs are exposed under `agent.readinessProbe.*`, following the same pattern as the existing `agent.livenessProbe.*`.

Resolves YOS-400.

# Pre-requisites

- [x] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [x] I have added a changeset with an appropriate customer facing description.
- [ ] This PR affects all release versions and will need to be forward merged.


charts/kubernetes-agent/tests/__snapshot__/tentacle-deployment_test.yaml.snap
```diff
@@ -123,6 +123,17 @@ should match snapshot:
                 periodSeconds: 30
                 timeoutSeconds: 5
               name: octopus-agent-tentacle
+              readinessProbe:
+                exec:
+                  command:
+                    - /bin/sh
+                    - -c
+                    - ps aux | grep -v grep | grep -q Tentacle || exit 1
```
liam-mackie: Nit: this may fail in certain scenarios (like in OpenShift) where there isn't permission to list processes, or grep doesn't exist. Since we've gone so long without having to do this, consider leaving the readiness check off by default.

liam-mackie (approved):
I know _why_ we're doing this, but it feels super weird to have a readiness check only on whether the main binary in it is running, which feels like something that the container runtime does for us for free anyway. 

I'd consider moving the readiness check to instead see if the tentacle has established a connection using `ss` or `lsof` if at all possible, or even by just watching a known directory for a file, if you have the time/inclination.

charts/kubernetes-agent/tests/__snapshot__/tentacle-deployment_test.yaml.snap
zentron: flipped around and disabled by default, considering its not actually particularly relevant.

